PR #6, opened 2026-03-17
fix: stream SSE error event when RAG generation step fails

dovinmu:
Generation errors (e.g. OpenAI quota exceeded) were logged server-side but never sent to the client, leaving the RAG Playground stuck on an empty placeholder. Now sends an SSEEventError so the frontend can display the failure.


lib/ai/errors_test.go
```diff
@@ -0,0 +1,145 @@
...
+	"context"
+	"errors"
+	"fmt"
+	"testing"
+
+	openrouter "github.com/revrost/go-openrouter"
+)
+
+func TestClassifyGenerationError_OpenRouterAuth(t *testing.T) {
+	err := &openrouter.APIError{HTTPStatusCode: 401, Message: "Invalid API key"}
```
ajroetker: This file could use some simplifications, maybe consolidate the tests to a single table based test?

lib/ai/errors.go
```diff
@@ -0,0 +1,151 @@
...
+		return 502
+	default:
+		return 500
+	}
+}
+
+// ClassifyGenerationError inspects the error chain for known provider SDK types
+// and returns a user-friendly message. The provider parameter is used to give
+// context in the message (e.g. "openrouter", "openai").
+func ClassifyGenerationError(provider string, err error) ClassifiedError {
```
ajroetker: Classify confused me a little bit because Classify means something different in `lib/ai` especially. Could we rename this to `AsGenerationError(` and `GenerationError` for the return type?

ajroetker (commented):
## Review

Good fix — generation errors were silently swallowed server-side, now they're properly surfaced to the client. The `ClassifyGenerationError` helper is well-structured with good test coverage.

### Issues

**1. Stale doc comment on `resolveProviderName`** (`retrieval_agent.go:1949`) — The doc comment for `convertQueryHitsToDocuments` got concatenated onto `resolveProviderName`:

```go
// convertQueryHitsToDocuments converts QueryHit slice to schema.Document slice
// resolveProviderName returns the provider name from the request for use in
// user-facing error messages.
func resolveProviderName(req *RetrievalAgentRequest) string {
```

The first line belongs on `convertQueryHitsToDocuments` below. Should remove it from above `resolveProviderName`.

**2. `isRateLimitError` overlap with `classifyByStatusCode`** — If the error is a typed SDK error with status 429, it's caught by the `errors.As` checks (steps 1-3) via `classifyByStatusCode`. The fallback `isRateLimitError` at step 5 only fires for untyped errors with rate-limit strings. Correct but non-obvious — worth a short comment explaining step 5 catches rate limits from non-SDK error types.

### Verdict

Fix the stale doc comment (#1), otherwise LGTM.

PR #46, opened 2026-04-16
fix(termite-sdk): bundle openapi-fetch to fix CJS interop crash

markhayden:
## Summary

Same root cause and fix as `@antfly/sdk` in #45 — applied to `@antfly/termite-sdk`.

`@antfly/termite-sdk` CJS bundle (`dist/index.cjs`) crashes on `new TermiteClient(...)` in any CJS consumer project loaded via tsx, ts-node, or plain `require()`:

```
TypeError: (0 , import_openapi_fetch.default) is not a function
    at new TermiteClient (dist/index.cjs:97:52)
```

## Root cause

`openapi-fetch@0.17` is ESM-first. When `tsup` emits the CJS bundle, it wraps the dependency as:

```js
var import_openapi_fetch = __toESM(require("openapi-fetch"), 1);
```

The `__toESM` interop wrapper double-nests the default export — `.default` resolves to a module object rather than the `createClient` function. The ESM build was unaffected.

## The fix

Introduce `ts/packages/termite-sdk/tsup.config.ts` with:

```ts
noExternal: ["openapi-fetch"],
```

This inlines `openapi-fetch` into both CJS and ESM outputs. No `require()` call, no `__toESM` wrapper, no interop to go wrong. The bug is eliminated by construction.

## Additional changes

- **`ts/packages/termite-sdk/package.json`**: `build` / `dev` scripts simplified to `tsup` / `tsup --watch`. `prepublishOnly` reordered (`build` before `test`) so the dist-guard test runs against a fresh build.
- **`ts/packages/termite-sdk/test/dist.test.ts`** (new): regression guard that loads `dist/index.cjs` after build and asserts:
  1. The broken `require("openapi-fetch")` / `import_openapi_fetch` patterns are absent.
  2. `new TermiteClient({...})` does not throw.

  **Proven to fail** when `noExternal` is removed. Skips gracefully when `dist/` is absent.

## Verification

Reproduced the crash against the unfixed build from current `main` (packed via `pnpm pack`, installed in a CJS consumer project, Node 22.13.1 + tsx 4.21.0). Exact `TypeError` observed at `dist/index.cjs:97:52`.

After the fix, the same harness passes. All 52 existing tests pass (50 pre-existing + 2 new guard tests), 1 expected skip. Typecheck clean.

## Test plan

- [x] Reproduce the original `TypeError` against unfixed build from `main`
- [x] Same consumer project passes against the fixed build
- [x] `pnpm --filter @antfly/termite-sdk test` — all existing tests still pass
- [x] Typecheck clean
- [x] `dist.test.ts` fails as designed when `noExternal` is removed (guard proven effective)
- [x] `dist.test.ts` skips cleanly when `dist/` is absent

### DIFF
--- a/ts/packages/termite-sdk/package.json
+++ b/ts/packages/termite-sdk/package.json
@@ -19,8 +19,8 @@
     "LICENSE"
   ],
   "scripts": {
-    "build": "tsup src/index.ts --format cjs,esm --dts --clean",
-    "dev": "tsup src/index.ts --format cjs,esm --dts --watch",
+    "build": "tsup",
+    "dev": "tsup --watch",
     "test": "vitest run",
     "test:watch": "vitest",
     "test:ui": "vitest --ui",
@@ -31,7 +31,7 @@
     "generate": "openapi-typescript ../../../termite/openapi.yaml -o ./src/termite-api.d.ts --default-non-nullable=false",
     "clean": "rm -rf dist",
     "prepare": "pnpm run build",
-    "prepublishOnly": "pnpm run typecheck && pnpm run lint && pnpm run test && pnpm run build"
+    "prepublishOnly": "pnpm run typecheck && pnpm run lint && pnpm run build && pnpm run test"
   },
   "keywords": [
     "termite",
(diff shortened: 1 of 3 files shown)


ajroetker:
I'm folding this one into the 45 and folding that into my other pr via a merge of your branch here

PR #85, opened 2026-05-20
Antfarm Zig Updates 

timkaye11:
## Summary

  Makes the Zig Antfly runtime the default Antfarm path and removes the old Cloudflare R2 registry
  assumptions from the Antfarm model experience. Antfarm now reads live model inventory from the
  embedded Termite `/ml/v1/models` API, while the Go binary remains available only through the explicit
  `build-go` target.

  ## Changes

  - Switches top-level `make build` to build the Zig `antfly` binary by default.
  - Keeps the legacy Go binary build behind `make build-go`.
  - Embeds/serves Antfarm from the Zig swarm runtime.
  - Adds Termite-compatible `/ml/v1` and `/termite/ml/v1` routing in the Zig runtime.
  - Updates Antfarm model loading to use live Termite model data instead of the old `/registry/
  index.json` registry.
  - Removes Antfarm-facing R2 registry proxy/config compatibility surface, including the deprecated
  `registry_url` config field.
  - Updates Antfarm UI model cards and playground links to work with live Termite model identifiers.
  - Gates MLX autodetection on required MLX symbols so default builds work on machines without a
  compatible local MLX install, while explicit `-Dmlx=true` fails early with a clear error.

  ## Validation

  - `make -C ./zig build`
  - `zig build lib-common-config-test`
  - `GOEXPERIMENT=simd go test ./src/common`
  - `make -C ./zig openapi-check`
  - `make -C ./zig build ZIG_BUILD_FLAGS=-Dmlx=true` verified the intentional early MLX compatibility
  failure path
  - Local Antfarm smoke-tested against the Zig `antfly` runtime, without Docker


ajroetker:
```
# github.com/antflydb/antfly/src/metadata
Error: src/metadata/antfarm.go:57:17: undefined: url
Error: src/metadata/antfarm.go:62:12: undefined: httputil
Error: src/metadata/antfarm.go:63:20: undefined: httputil
Error: src/metadata/antfarm.go:65:16: undefined: strings
Error: src/metadata/antfarm.go:69:21: undefined: path
Error: src/metadata/antfarm.go:78:32: undefined: time
Error: Process completed with exit code 1.
``` 
real failures

zig/pkg/antfly/src/swarm/runtime.zig
```diff
@@ -834,6 +843,9 @@ fn serveUnifiedInner(
     // Register termite routes under /ml/v1
     if (termite_node) |node| {
         try node.registerRoutesOn(termite.server.public_api_prefix, &server);
+        try node.registerRoutesOn("/termite" ++ termite.server.public_api_prefix, &server);
```
ajroetker: Why are we registering these on /termite? Termite shares a health server at / and /ml/v1 should be accessible on the main server?

zig/pkg/antfly/src/swarm/runtime.zig
```diff
@@ -893,6 +906,163 @@ fn readyzHandler(ctx: *httpx.Context) anyerror!httpx.Response {
...
+}
+
+fn isAntfarmReservedPath(path: []const u8) bool {
+    const reserved = [_][]const u8{
+        "/api",
+        "/ml",
+        "/termite",
+        "/metadata",
+        "/internal",
+        "/_internal",
```
ajroetker: I think all the _internal got moved to internal or vice versa but can we consolidate to the zig pathing? I thought i did in go already?

zig/pkg/termite/build.zig
```diff
@@ -53,14 +53,51 @@ fn defaultOnnxRuntimeRoot(b: *std.Build, target: std.Build.ResolvedTarget) []con
...
+    ) catch return false;
+    defer b.allocator.free(bytes);
+
+    for (required_mlx_symbols) |symbol| {
+        if (std.mem.indexOf(u8, bytes, symbol) == null) return false;
+    }
+    return true;
+}
+
+fn mlxRootAvailable(b: *std.Build, target: std.Build.ResolvedTarget, root: []const u8) bool {
```
ajroetker, line 56: fwiw i turned off mlx and onnx binding by default, don't know if that's applicable here

Makefile
```diff
@@ -32,9 +33,10 @@ help:
 	@echo "Usage: make <target>"
 	@echo ""
 	@echo "Targets:"
-	@echo "  build              Build the antfly binary"
+	@echo "  build              Build the Zig antfly binary"
+	@echo "  build-go           Build the legacy Go antfly binary"
 	@echo "  build-antfarm      Build the antfarm frontend (React admin UI)"
-	@echo "  build-termite-dashboard  Build the termite dashboard (termite-only antfarm)"
+	@echo "  build-termite-dashboard  Build the legacy Go termite dashboard (termite-only antfarm)"
```
ajroetker: Can we just remove this?

ajroetker: Probably should remove the legacy standalone termite target altogether if that simplifies the dashboard?

zig/pkg/antfly/src/swarm/runtime.zig
timkaye11: fixed

zig/pkg/antfly/src/swarm/runtime.zig
timkaye11: cleaned this up on the zig path

zig/pkg/termite/build.zig
timkaye11: Updated both Zig build surfaces so MLX and ONNX are opt-in instead of auto-enabled

PR #88, opened 2026-05-20
Harden store backup path handling

dovinmu:
## Summary
- validate backup IDs before using them in filesystem paths
- keep local `file://` backup and restore paths scoped under the Antfly base directory
- use safe path joining for shard backup/restore paths and S3 destination keys
- add regression coverage for traversal backup IDs and out-of-scope local backup locations

## Testing
- `go test ./src/common ./src/store -run 'TestValidateBackupID|TestHandleBackup|TestHandleStartShard' -count=1 -timeout=60s`
- `git diff --check`

---

*Generated using snyk + codex*


ajroetker:
e2e test failures are probably real

PR #95, opened 2026-05-21
Jina v5 direct metal embeddings

timkaye11:
## Summary

Adds a resident direct-Metal graph execution path for Jina v5 / Qwen3-style text embeddings. Eligible embedding models now run token lookup, dense Qwen3 layers, final norm, pooling, and normalization on Metal, materializing only final vectors to host.

## What Changed

- Added `qwen3_dense_text_embedding` graph-frame contract.
- Added Jina/Qwen3 embedding pipeline path:
  - graph-frame Metal execution
  - resident Metal fallback
  - existing `session.run` fallback
- Added dense causal no-KV Qwen3 execution on direct Metal.
- Added batched head RMSNorm + RoPE with per-sequence position reset.
- Prepared resident slots for Qwen3/Jina weights, including LoRA-merged Jina retrieval adapter weights.
- Added resident pooling for flattened `[batch * seq, hidden]` outputs.
- Manifest-gated the path to Jina/Qwen3 embedding models only.
- Removed development-only Qwen3 env toggles and unused graph fields.

## Validation

- `zig fmt`
- `git diff --check`
- `zig build termite-test -Dmlx=false -Dmetal=true -Donnx=false -Dcuda=false`
- Real Jina v5 Metal smoke:
  - batch 1 graph success
  - batch 8 graph success
  - graph path confirmed via `text.encoder.qwen3.graph`
- Observed graph encoder timing:
  - batch 1 ~`216ms`
  - batch 8 ~`660-730ms`

## Notes

Scoped to direct Metal dense Jina/Qwen3 embedding manifests. Qwen3.5, MoE, PLE, multimodal, sparse, MLX, ONNX, CUDA, and generic Qwen3 generation remain on existing paths.

### DIFF
--- a/zig/pkg/termite/src/api/openapi.yaml
+++ b/zig/pkg/termite/src/api/openapi.yaml
@@ -1121,14 +1121,29 @@ components:
         dimensions:
           type: integer
           description: Optional truncation size for dense embeddings. Must be a positive integer no larger than the model embedding size. Not supported for sparse models.
+        task_type:
+          type: string
+          description: Optional embedding task type using Google embedding task-type names. For Jina v5 text embeddings, query-side tasks use the query prefix and RETRIEVAL_DOCUMENT uses the document prefix.
+          enum:
+            - RETRIEVAL_QUERY
+            - RETRIEVAL_DOCUMENT
+            - QUESTION_ANSWERING
+            - FACT_VERIFICATION
+            - CODE_RETRIEVAL_QUERY
+            - CLASSIFICATION
+            - CLUSTERING
+            - SEMANTIC_SIMILARITY
         input_type:
           type: string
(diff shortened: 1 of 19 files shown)


zig/pkg/termite/src/api/openapi.yaml
```diff
@@ -1121,6 +1121,15 @@ components:
         dimensions:
           type: integer
           description: Optional truncation size for dense embeddings. Must be a positive integer no larger than the model embedding size. Not supported for sparse models.
+        task:
```
ajroetker: Is there a cohere, jina or openai standard format for this?

ajroetker:
Sorry tim https://docs.cloud.google.com/gemini-enterprise-agent-platform/models/embeddings/task-types I'm wondering if we should standardize on google instead of cohere since we use openai for everything else

PR #107, opened 2026-05-23
Bedrock Titan Multi-modal Support

excenter:
and tests


excenter:
the last commit was just to make lint happy. :/

Passes my local tests, but would probably need a review by someone who deeply understands the code I copy-pasted from elsewhere.

excenter:
#104 looks better in almost every way. feel free to discard this. 
The only thing I see from my PR worth keeping over that is treating imageURL as a source for the image to be downloaded and embedded vs treating it as text.

ajroetker:
I merged this into my other branch so you get the credit in the git history @excenter! Thanks for taking a look at this! I worked on getting the zig version working for this too for you

PR #121, opened 2026-05-26
Fix TTL metadata for local document writes

### LINKED ISSUE #120: TTL cleanup job runs but never expires documents (swarm mode, v0.1.1)
[H-05-v011-antfly.log](https://github.com/user-attachments/files/28236983/H-05-v011-antfly.log)

[H-05-v011-repro.sh](https://github.com/user-attachments/files/28236985/H-05-v011-repro.sh)

## Summary

On AntFly v0.1.1, a table configured with `schema.ttl_duration` starts a TTL cleanup worker but never expires any document. Records remain readable indefinitely via `GET /api/v1/tables/{name}/lookup/{key}` and the worker reports `total_documents_expired=0` on shutdown, even when the inserted records are far past their expiration window (default `_timestamp` case) or use a stale custom `ttl_field` value (six years past expiration).

This contradicts both the documented behavior (`docs/ttl-example.md` says "Get operations return 'not found' for expired documents" and that a background job "deletes them every 30 seconds") and what the worker logs say it is doing.

## Environment

- AntFly version: `0.1.1` (binary `antfly-omni_0.1.1_Darwin_arm64.tar.gz` from `releases.antfly.io`)
- OS: macOS 26.5, arm64 (Apple Silicon)
- Mode: single-node `antfly swarm --termite=false`
- Storage: fresh `HOME` so no prior state contaminates the test

This may be specific to swarm mode / `--termite=false` / the local-shard-bypass code path — happy to retest with a different topology if helpful.

### Setup — start a disposable node

```bash
mkdir -p /tmp/antfly-repro && HOME=/tmp/antfly-repro \
  antfly swarm --termite=false \
  --metadata-api http://127.0.0.1:18082 \
  --metadata-raft http://127.0.0.1:19018 \
  --store-api    http://127.0.0.1:12382 \
  --store-raft   http://127.0.0.1:19022
```

### Case A — default `_timestamp` TTL

Create the table (note: `ttl_duration` must go inside `schema`; see "Documentation also wrong" below):

```bash
curl -s -X POST http://127.0.0.1:18082/api/v1/tables/ttl_default \
  -H "Content-Type: application/json" \
  -d '{
    "num_shards": 1,
    "schema": {
      "ttl_duration": "2s",
      "document_schemas": {
        "memory_entry": {
          "schema": {
            "type": "object",
            "properties": {
              "summary_text": {"type": "string", "x-antfly-types": ["text"]}
            }
          }
        }
      }
    }
  }'
```

Worker starts (good):

```
caller=db/ttl.go:83 msg="Starting TTL cleanup job" ttl_duration=2s cleanup_interval=30s grace_period=5s
```

Insert one record:

```bash
curl -s -X POST http://127.0.0.1:18082/api/v1/tables/ttl_default/batch \
  -H "Content-Type: application/json" \
  -d '{"inserts":{"ts-001":{"_type":"memory_entry","summary_text":"default _timestamp test"}}}'
```

Wait > `ttl_duration + cleanup_interval + grace_period` (= 2s + 30s + 5s = 37s). I waited 3+ minutes across two cleanup cycles.  I tried multiple times using, with longer wait periods. Lookup:

```bash
sleep 40s; curl -s -w "\nHTTP %{http_code}\n" \
  http://127.0.0.1:18082/api/v1/tables/ttl_default/lookup/ts-001;
sleep 40s; curl -s -w "\nHTTP %{http_code}\n" \
  http://127.0.0.1:18082/api/v1/tables/ttl_default/lookup/ts-001
```

**Expected:** record is filtered out at GET time (per docs); cleanup job logs a per-cycle "Cleaned up expired documents count=…" entry.

**Actual:** record still returned with HTTP 200:

```json
{"_timestamp":"2026-05-25T18:31:19.237684Z","_type":"memory_entry","summary_text":"default _timestamp test"}
```

On shutdown:

```
caller=db/ttl.go:94 msg="TTL cleanup job stopped" total_documents_expired=0
```

### Case B — custom `ttl_field`, six-year-stale record

Create:

```bash
curl -s -X POST http://127.0.0.1:18082/api/v1/tables/ttl_field \
  -H "Content-Type: application/json" \
  -d '{
    "num_shards": 1,
    "schema": {
      "ttl_duration": "2s",
      "ttl_field": "written_at",
      "document_schemas": {
        "memory_entry": {
          "schema": {
            "type": "object",
            "properties": {
              "written_at": {"type": "string", "format": "date-time"},
              "summary_text": {"type": "string", "x-antfly-types": ["text"]}
            },
            "required": ["written_at"]
          }
        }
      }
    }
  }'
```

Insert a record that is *already* six years past expiration:

```bash
curl -s -X POST http://127.0.0.1:18082/api/v1/tables/ttl_field/batch \
  -H "Content-Type: application/json" \
  -d '{"inserts":{"old-001":{"_type":"memory_entry","written_at":"2020-01-01T00:00:00Z","summary_text":"stale"}}}'
```

Wait one cleanup cycle (≥ 35 s) and look up:

```bash
sleep 40; curl -s http://127.0.0.1:18082/api/v1/tables/ttl_field/lookup/old-001
```

**Actual:** record still returned (HTTP 200).

On shutdown:

```
caller=db/ttl.go:94 msg="TTL cleanup job stopped" total_documents_expired=0
```

## What the logs do show

The TTL workers started, ran the full session, and stopped cleanly — but never produced any per-cycle cleanup log entry, and the final `total_documents_expired` was 0 for both shards. No errors, no panics, nothing in the log explains why nothing expired.

```
Starting TTL cleanup job ... ttl_duration=2s cleanup_interval=30s grace_period=5s
... (no cleanup-cycle log lines across ~4 minutes) ...
TTL cleanup job stopped ... total_documents_expired=0
```

A full session log is available — happy to attach.

## Plausible root cause (speculative)

I haven't dug into the code, but two guesses worth checking:

1. **Local-shard-bypass write path doesn't write the `:t` side key.** The docs describe a separate `:t` key that holds the raw TTL timestamp for fast scans. If swarm mode's local bypass inserts the document but skips writing `:t`, then `cleanupExpiredDocuments` would iterate without finding anything, and the real-time GET filter would also see no `:t` key and return the doc unfiltered. This would match all observed behavior (worker starts, no errors, zero expired, GETs return expired docs).
2. **Cleanup ticker fires on a clock that isn't advancing in this mode.** v0.1.1 refactored `TTLCleaner` to use an injectable `clock.Clock`. If something in swarm mode wires a non-advancing clock, the ticker would never fire — but you'd also expect zero log lines, which matches. Less likely but worth a glance.

I have no opinion on which (if either) is right; the maintainers will know much faster than I will.

## Documentation also wrong (separate, small)

While reproducing this I hit two docs issues. Both are independent of the bug above; flagging in case they're useful:

1. `docs/ttl-example.md` shows `ttl_duration` and `document_schemas` at the **top level** of the create-table body, but the actual `Table` schema in `openapi.yaml` puts both inside `schema` (`TableSchema`). The doc shape is silently accepted by the API but the TTL worker never starts — easy to mistake for the bug above.
2. `docs/ttl-example.md` shows `POST /api/v1/tables` (no path arg), but the actual endpoint per `openapi.yaml` is `POST /api/v1/tables/{tableName}`. The doc shape returns `Method Not Allowed`.

## What changed between v0.1.0 and v0.1.1 for `src/store/db/ttl.go`

Diffing `v0.1.0…v0.1.1` for that file shows only refactoring: introducing a `clock.Clock` abstraction for testability, switching `recover()` to a shared `pebbleutils.RecoverPebbleClosed`, switching `tc.db.byteRange` to a `getByteRange()` getter, and replacing a raw `indexManager.Batch` call with `tc.db.BatchIndex`. Nothing in that diff looks like it would change observable expiration behavior, which is consistent with this bug being present on both versions.

## What I'd find useful in a fix or a "won't fix, but here's why"

- Confirmation of whether swarm mode + local-shard-bypass is supposed to participate in TTL at all (vs. it being a known limitation of the single-node `--termite=false` setup).
- If TTL is supposed to work here, a small repro test in `src/store/db/ttl_test.go` that exercises the end-to-end insert → wait → GET path, not just the cleanup logic in isolation.

Thanks for the work on AntFly — happy to do additional repros against a multi-node config or a debug build if it helps.

ajroetker:
Fixes #120.

## Summary
- write TTL `:t` metadata for TTL-enabled document writes even when the context does not carry a timestamp
- derive document TTL metadata from custom `ttl_field` values when present
- normalize document timestamp side-key lookup/cleanup so query filtering and TTL cleanup use the same key shape as writes, with fallback support for legacy `key:t` metadata
- add regressions for local writes without context timestamps and custom TTL fields

## Tests
- `env GOWORK=off go test ./src/store/db -run 'TestBatchWritesTTLTimestampsWithoutContextTimestamp|TestBatchDerivesTTLTimestampsFromCustomField|TestTTLCleanup|TestTTLQueryFiltering|TestTTLDocumentUpdateTimestamp'`
- `env GOWORK=off go test ./src/store/db`
- `git diff --check`

### DIFF
--- a/go/pkg/antfly/src/store/db/db.go
+++ b/go/pkg/antfly/src/store/db/db.go
@@ -2841,6 +2841,73 @@ func shouldWriteTimestamp(key []byte) bool {
 	return true
 }
 
+func (db *DBImpl) defaultWriteTimestamp(timestamp uint64) uint64 {
+	if timestamp != 0 {
+		return timestamp
+	}
+	if db.cachedTTLDuration == 0 {
+		return 0
+	}
+	return uint64(time.Now().UTC().UnixNano()) //nolint:gosec // UnixNano is positive for supported wall-clock dates.
+}
+
+func (db *DBImpl) resolveWriteTimestamp(docJSON []byte, fallback uint64) uint64 {
+	if db.schema == nil || db.cachedTTLDuration == 0 {
+		return fallback
+	}
+
+	var doc map[string]any
+	if err := json.Unmarshal(docJSON, &doc); err != nil {
+		db.logger.Warn("could not parse document for TTL timestamp",
+			zap.Error(err))
+		return fallback
+	}
+
+	value, ok := doc[db.schema.GetTTLField()]
+	if !ok {
+		return fallback
+	}
+
+	timestamp, ok := ttlTimestampFromValue(value)
+	if !ok {
+		db.logger.Warn("could not parse TTL field timestamp",
+			zap.String("ttl_field", db.schema.GetTTLField()))
+		return fallback
+	}
+	return timestamp
+}
+
+func ttlTimestampFromValue(value any) (uint64, bool) {
+	switch v := value.(type) {
+	case string:
+		if ts, err := time.Parse(time.RFC3339Nano, v); err == nil {
+			return unixNanoToUint64(ts)
+		}
+		if ts, err := time.Parse(time.RFC3339, v); err == nil {
+			return unixNanoToUint64(ts)
+		}
+	case float64:
+		if v >= 0 {
+			return uint64(v), true
+		}
+	case int64:
+		if v >= 0 {
+			return uint64(v), true
+		}
+	case uint64:
+		return v, true
+	}
+	return 0, false
+}
+
+func unixNanoToUint64(ts time.Time) (uint64, bool) {
+	nanos := ts.UnixNano()
+	if nanos < 0 {
+		return 0, false
+	}
+	return uint64(nanos), true //nolint:gosec // negative values are handled above.
+}
+
 func (db *DBImpl) Batch(
 	ctx context.Context,
 	writes [][2][]byte,
@@ -2859,8 +2926,10 @@ func (db *DBImpl) Batch(
 		return pebble.ErrClosed
 	}
 
-	// Get HLC timestamp from context (set by metadata layer for consistent timestamps)
-	timestamp := storeutils.GetTimestampFromContext(ctx)
+	// Get HLC timestamp from context (set by metadata layer for consistent timestamps).
+	// Single-node/local apply paths may not carry one; TTL still needs a side-key
+	// timestamp so query-time filtering and cleanup can see new writes.
+	timestamp := db.defaultWriteTimestamp(storeutils.GetTimestampFromContext(ctx))
 	splitState := db.splitState
 	mergeState := db.mergeState
 
@@ -3019,7 +3088,9 @@ func (db *DBImpl) Batch(
 		valueToCompress := kv[1]
 		shouldWriteDocument := true
 
-		sfResult, err := db.extractAndWriteSpecialFields(batch, kv[1], kv[0], timestamp)
+		writeTimestamp := db.resolveWriteTimestamp(kv[1], timestamp)
+
+		sfResult, err := db.extractAndWriteSpecialFields(batch, kv[1], kv[0], writeTimestamp)
 		if err != nil {
 			return fmt.Errorf("extracting special fields for key %s: %w", kv[0], err)
 		}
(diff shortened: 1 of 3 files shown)


ajroetker:
zig already has tests for this

PR #131, opened 2026-05-28
Fix bool must+filter clause handling in structured filter parser

dovinmu:
### Motivation

- The structured-filter parser treated the presence of `filter` as if it were part of the required (`must`) clauses but then only parsed `filter` when `must` was absent, which caused `bool` queries containing both `must` and `filter` to drop the `filter` clauses and allowed under-enforcement of row-level policies.

### Description

- In `zig/pkg/antfly/src/storage/db/query/search_exec.zig`, the converter now builds one required clause list by appending parsed `filter` entries followed by parsed `must` entries.
- The conversion path now appends directly into the output list, avoiding the intermediate slice leak from the first revision.
- The sibling structured-filter paths in `zig/pkg/antfly/src/storage/db/query/graph_exec.zig` and `zig/pkg/antfly/src/search/pattern_filter.zig` now enforce `bool.filter` clauses as required clauses instead of ignoring them.
- The C API text query parser in `zig/pkg/antfly/src/capi/db.zig` now merges `bool.filter` into required clauses.
- Replication route filter normalization in `zig/pkg/antfly/src/metadata/replication_backfill.zig` now preserves `bool.filter` clauses.
- Pattern-filter text-stat collection now also walks `bool.filter` clauses.
- Planning stats now treats `bool.filter` as a required structured-filter clause for clause counts, exact-count eligibility, and positive doc-id upper bounds.
- Algebraic constraint collection now includes `bool.filter` as a required clause when resolving doc-id candidate sets through the constraint fast path.
- Existing `should`, `must_not`, `only_must_not`, and `min_should` behavior is preserved.

### Testing

- `zig build root-test -- "metadata replication route bool filter clauses are required with must clauses" "pattern bool filter clauses are merged into required search clauses" "jsonDocMatchesPatternFilter supports stored structured filters" "stored doc bool filter clauses are required with must clauses" "planning stats treats bool filter clauses as required structured filters" "algebraic doc facts expose doc id candidate sets for symbolic constraints"`
- `zig build capi-test -- "c api bool text parser treats filter clauses as required"`
- `zig build root-test`
- `zig build capi-test`

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a164877b69c8321bde4bbe5e5fc7cc5)

### DIFF
--- a/zig/pkg/antfly/src/capi/db.zig
+++ b/zig/pkg/antfly/src/capi/db.zig
@@ -3841,8 +3841,16 @@ fn parseTextQueryJson(alloc: std.mem.Allocator, value: std.json.Value) anyerror!
     if (value.object.get("bool")) |bool_query| {
         if (bool_query != .object) return error.InvalidArgument;
 
-        const must = if (bool_query.object.get("must")) |must_value|
-            try parseTextQueryArrayJson(alloc, must_value)
+        var must_list = std.ArrayListUnmanaged(db_mod.types.TextQuery).empty;
+        errdefer must_list.deinit(alloc);
+        if (bool_query.object.get("filter")) |filter_value| {
+            try appendTextQueryArrayJson(alloc, &must_list, filter_value);
+        }
+        if (bool_query.object.get("must")) |must_value| {
+            try appendTextQueryArrayJson(alloc, &must_list, must_value);
+        }
+        const must = if (must_list.items.len > 0)
+            try must_list.toOwnedSlice(alloc)
         else
             &.{};
         const should = if (bool_query.object.get("should")) |should_value|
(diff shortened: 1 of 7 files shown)


ajroetker:
There's a flaky e2e test on linux I'm trying to fix in my other PR, this lgtm

ajroetker:
```• Findings

  - High: bool filter is still silently dropped in other structured-filter evaluators.
    PR #131 fixes patternBoolFilterToSearchQuery, but the same JSON shape still goes through graph_exec.compilePatternFilter via /private/tmp/antfly-pr131/zig/pkg/antfly/src/storage/db/query/search_exec.zig:1518. That compiler only reads must,
    should, and must_not at /private/tmp/antfly-pr131/zig/pkg/antfly/src/storage/db/query/graph_exec.zig:1593, so {"bool":{"must":[...],"filter":[...]}} still compiles and enforces only must. The direct JSON matcher has the same gap at /private/
    tmp/antfly-pr131/zig/pkg/antfly/src/search/pattern_filter.zig:55. For a security fix, all structured-filter execution paths should either enforce filter as required clauses or reject it consistently.
  - Medium: the new merge path leaks the intermediate slices returned by patternFilterArrayToSearchQueries.
    At /private/tmp/antfly-pr131/zig/pkg/antfly/src/storage/db/query/search_exec.zig:2669, filter_queries is allocated, shallow-copied into must_out, then never freed. Same for must_queries at line 2673. Production often uses an arena here, but
    the ownership is still wrong and will show up under non-arena/testing allocators. Prefer building directly into one list or freeing only the temporary slice backing after ownership of the elements is transferred.
```

PR #139, opened 2026-05-29
Add Antfly Cloud CLI shim

dovinmu:
## Summary
- add `antfly cloud ...` to the Zig Antfly CLI as a thin shim that delegates to `antfly-cloud` on PATH
- keep the same shim in the legacy Go CLI for low-cost compatibility after the Go tree moved under `go/pkg/antfly`
- print explicit install/help text when `antfly-cloud` is missing
- preserve child process stdio/env/args and propagate child exit codes
- avoid vendoring or coupling the Antfly Cloud CLI into either Antfly CLI

## Validation
- `zig build antfly-main-test`
- `make zig-build`
- `(cd go/pkg/antfly && go test ./cmd/...)`
- built Zig `antfly` locally and verified missing-binary path exits 127 with install instructions
- built Zig `antfly` locally and verified fake `antfly-cloud` receives args/stdin/env and exit code is propagated
- built legacy Go `antfly` locally and verified missing-binary and fake `antfly-cloud` delegation paths


go/pkg/antfly/cmd/cmd/cloud.go
```diff
@@ -0,0 +1,72 @@
...
+	"errors"
+	"fmt"
+	"io"
+	"os"
+	"os/exec"
+
+	"github.com/spf13/cobra"
+)
+
+const antflyCloudBinary = "antfly-cloud"
```
ajroetker, line 13: We need this on the zig version as well

PR #164, opened 2026-06-04
potential bugfixes from data loading stress test on linux

dovinmu:
[WIP]


ajroetker:
fixed in my other PR, at least the data.zig changes, I didn't keep the build changes because they're a bit far reaching

Issue #205, opened 2026-06-07
main: antflydb/clipclap text embeddings are degenerate via /ai/v1/embed — ~96% shared component, length-organized geometry, cross-modal scores collapse to a 0.22–0.24 band


markhayden:
## Environment

antfly built from main @ 83492927d, swarm mode, darwin-arm64. Model: `antflydb/clipclap` pulled via `antfly inference pull antflydb/clipclap:gguf:Q4_K --capabilities image,audio`, loaded on metal. Embeddings via `POST /ai/v1/embed` directly — no enrichment pipeline involved.

## Symptom

Text embeddings from clipclap are degenerate: retrieval over them is noise. Measured on a clean instance:

**1. No semantic discrimination — geometry organizes by text length, not meaning:**

```
'cat' ~ 'kitten'                                          0.868
'cat' ~ 'quarterly financial report'                      0.839   ← cross-topic
'cat' ~ 'a fluffy young cat playing with yarn in the sun' 0.818   ← SAME topic, scores LOWER
'quarterly financial report' ~ 'spreadsheet with quarterly revenue figures...' 0.937
'kitten' ~ 'spreadsheet with quarterly revenue figures...' 0.862
```

Every pair lands in 0.82–0.94 regardless of topic; short~short pairs beat short~long same-topic pairs; the long~long pair tops everything.

**2. ~96% of every text vector is a shared constant component.** Projecting each vector onto the corpus-mean direction explains a mean fraction of **0.959** of its norm (per-text 0.90–0.99). The actual semantic signal lives in the residual ~4%.

**3. Cross-modal scores collapse.** Against an embedded image of a plain red square, all texts score in a 0.224–0.235 band — and `'kitten'` (0.2352) **out-scores the literal description** `'a red square'` (0.2348).

**4. Control: BGE through the identical harness discriminates fine** (`cat`~`kitten` 0.964 vs `kitten`~`spreadsheet…` 0.566; spread 0.57–0.96, ordered by relatedness; shared-component 0.89). So this is a clipclap serving issue, not the harness.

## Repro

Embed the seven texts above plus one image via `/ai/v1/embed` (`{"model":"antflydb/clipclap","input":"<text>"}`; image as `{"type":"media","data":"<b64>","mime_type":"image/png"}`), compute pairwise cosines and the corpus-mean projection ratio. 30-second pass/fail probes:

- same-topic short~long must beat cross-topic short~short (currently fails)
[...]

Issue #208, opened 2026-06-07
main: TERMITE_PREFERRED_BACKEND=onnx hard-fails GGUF-only models with NoModelFileFound — no fallback, misleading error


markhayden:
## Environment

antfly built from main @ 83492927d, swarm mode, darwin-arm64. Model: `antflydb/clipclap` (GGUF-only package).

## Symptom

With `TERMITE_PREFERRED_BACKEND=onnx` set, GGUF-only models hard-fail instead of falling back to a backend that can load them:

```bash
TERMITE_PREFERRED_BACKEND=onnx antfly swarm ...
curl -X POST localhost:5738/ai/v1/embed -H 'Content-Type: application/json' \
  -d '{"model":"antflydb/clipclap","input":"hello"}'
# -> 500 {"error":"MODEL_LOAD_FAILED","message":"NoModelFileFound"}
```

```
trying backend onnx for ~/.antfly/inference/models/antflydb/clipclap
loadModel(...) failed: no backend accepted model
```

Unpinned, the same model loads fine on metal. So the "preferred" backend semantics are actually exclusive: a preference bricks a loadable model.

## Why this bites

The env var is the practical workaround for backend-specific issues (e.g. pinning ONNX while Metal inference had stability problems pre-#198) — but applying it node-wide silently breaks every GGUF-only model on the node, and `NoModelFileFound` points at missing files rather than the pin.

## Ask

Either (a) treat the pin as a *preference*: try the preferred backend first, fall back to others that accept the model (possibly with a log line), or (b) keep it strict but make the error say so — e.g. `BackendPinExcludedModel: model has no onnx artifacts and TERMITE_PREFERRED_BACKEND=onnx is set`. Related: is `TERMITE_PREFERRED_BACKEND` a supported/stable knob, or an internal that may vanish?

Issue #207, opened 2026-06-07
main: one failing embeddings index still marks healthy sibling indexes failed (worker-scoped failure state) — #191 blast radius persists post-#198


markhayden:
## Environment

antfly built from main @ 83492927d, swarm mode, darwin-arm64, single node.

## Symptom

One failing embeddings index still takes down every index on its table. With two embeddings indexes on one table — `mtext` (BGE over `{{title}}`, healthy config) and `mvisual` (media template, currently failing upstream for unrelated reasons, see #201):

```
mtext:   backfill_state=failed  total_indexed=1   ← it actually embedded its doc
mvisual: backfill_state=failed  total_indexed=0   ← the actual failure
```

The healthy index does its work (`total_indexed` advances) and is then marked `failed` anyway, with `backfill_active` stuck `true`; future enrichment for the whole table is dead.

## Mechanism

Same as reported in #191's blast-radius section: the enrichment failure state is **worker-scoped (per table), not index-scoped** — an error from one index's enrichment propagates out of the shared worker loop and `recordError` sets the single `worker_failed` flag (`storage/db/enrichment/enrichment_runtime.zig`, ~:1196), flooring every index on the table. #198 fixed #191's trigger but the isolation gap remains and any per-index failure (model issues, remote provider outages, bad templates) reproduces it.

## Expected

Per-index failure isolation: a failing index goes `failed`; healthy siblings keep enriching and report their true state.

Issue #206, opened 2026-06-07
main: fused hits never serialize _index_scores — fusion computes them, the response path drops them (spec promises the field; the Go runtime delivered it)


markhayden:
## Environment

antfly built from main @ 83492927d, swarm mode, darwin-arm64.

## Symptom

Fused (hybrid/RRF) hits never carry per-index scores — `_index_scores` is `null` on every hit, always:

```bash
curl -X POST localhost:5738/db/v1/tables/t/query -H 'Content-Type: application/json' \
  -d '{"semantic_search":"hello","indexes":["emb"],"full_text_search":{"query":"hello"},"limit":3}'
# hits: [{"_id":"d1","_score":0.0328,"_index_scores":null,...}, ...]
```

## Verification

The data is computed and then dropped on the floor:

- `zig/pkg/antfly/src/search/fusion.zig` — `FusionHit` carries `index_scores: []IndexScore` (~:61) and `rrfFuse` populates it per hit per source index (~:132-135).
- `zig/pkg/antfly/src/search/search.zig` `executeHybrid` (~:2198-2213) converts `FusionHit` → `ScoredHit{doc_id, score, id, stored_data}` — `ScoredHit` (~:422) has no field for index scores, so they're discarded; the response field defaults to `null`.
- The OpenAPI spec promises the field (openapi.yaml ~:9552): `_index_scores: ... "Scores partitioned by index when using RRF search."`
- The Go runtime delivered it: `go/pkg/antfly/src/store/db/indexes/remoteindex.go:199-206` populates `IndexScores` per hit and `metadata/api_query.go:1247-1250` threads it into the response. Migration regression.

## Why it's worth fixing beyond completeness

Per-leg scores are the only way to see that one leg of a hybrid query is contributing garbage. We tracked a multimodal-quality bug (#205) that stayed invisible for exactly this reason: RRF rankings looked plausible while one leg was noise, and every debug surface rendered the per-index contribution as 0.0/null.

PR #209, opened 2026-06-07
Fix inference and enrichment issue regressions

ajroetker:
## Summary

Fixes the open issue batch #201 through #208:

- #201: route local Antfly multimodal embedding through an in-process parts vtable and keep binary media base64 buffers alive until request serialization.
- #202: probe declared embedding model inputs before sending `attention_mask` or `token_type_ids`.
- #203: resolve symbolic tensor-view shapes during native materialization so reranker transpose paths do not receive empty tensors.
- #204: guard text segment stats slices against stale/invalid section offsets.
- #205: pool CLIP text embeddings at the EOT token position by highest token id.
- #206: preserve fused hit per-index scores through search, DB result materialization, and query response serialization; clean up partially initialized fused hits on conversion failure.
- #207: isolate non-retryable enrichment request failures so one bad index does not set worker-scoped failure for healthy siblings, while the failing index is still surfaced as failed in index status.
- #208: treat `ANTFLY_INFERENCE_PREFERRED_BACKEND` (and legacy `TERMITE_PREFERRED_BACKEND`) as a preference with fallback backends, including five-backend CUDA preference lists.

## Test Coverage

Added focused regression coverage for:

- local Antfly media parts routing and binary base64 request serialization
- fused `_index_scores` JSON serialization and DB result ownership cleanup on conversion failure paths
- isolated enrichment request errors leaving `worker_failed=false` and marking only the failed index
- index status encoding that reports one failed embeddings index without poisoning a healthy sibling
- preferred backend fallback ordering, including five-backend preference capacity
- declared text embedding input arity for `input_ids`, `attention_mask`, and `token_type_ids`
- CLIP text EOT pooling, including repeated highest-token padding
- invalid inverted section slice guards in segment stats and global field-length aggregation
- symbolic tensor-view materialization for transpose/rerank graph paths

## Verification

- `zig build`
- `zig build inference-test`
- `zig build lib-db-query-test`
- `zig build index-manager-test`
- `zig build inference-test lib-db-query-test lib-db-enrichment-worker-test lib-swarm-runtime-test index-manager-test`
- `zig build unit-test` partially run until the unrelated `data.storage.db_split_handoff.test.db merge coordinator opt-in applies configured receiver identity namespace` failure; the new `api.indexes` status test and enrichment runtime test passed before that failure.
- `git diff --check`

Warnings printed during inference/index-manager tests are from existing diagnostic/negative-path tests; commands exited successfully.

### DIFF
--- a/zig/build.zig
+++ b/zig/build.zig
@@ -3314,6 +3314,7 @@ pub fn build(b: *std.Build) void {
             "api table reads reject stale doc identity before multigroup fanout",
             "distributed table reads reject stale doc identity before multigroup fanout",
             "api public table query rejects only top-level internal fields",
+            "single embeddings index encoder scopes isolated enrichment failure to one index",
             "api query contract rejects doc identity control fields when with relaxes schema",
             "api query contract public parser rejects internal shard doc identity controls",
             "api distributed graph hydrate carries identity generation and clears cross-range ordinals",
@@ -3516,6 +3517,7 @@ pub fn build(b: *std.Build) void {
         .root_module = api_table_reads_docid_test_mod,
         .filters = &.{
             "provisioned read cache invalidates repeated ownership moves with pinned leases",
+            "parseRemoteSearchResult preserves fused index scores",
         },
         .test_runner = .{
             .path = b.path("pkg/antfly/src/test_runner.zig"),
--- a/zig/pkg/antfly/src/api/indexes.zig
+++ b/zig/pkg/antfly/src/api/indexes.zig
@@ -611,6 +611,7 @@ const AggregatedIndexStatus = struct {
     kind: ?db_mod.types.IndexKind = null,
     backfill_active: bool = false,
     backfill_progress: f64 = 0.0,
+    enrichment_failed: bool = false,
     table_doc_count: u64 = 0,
     doc_count: u64 = 0,
     term_count: u64 = 0,
@@ -745,6 +746,7 @@ fn aggregateIndexStatus(
         aggregate.replay_applied_sequence += item.replay_applied_sequence;
         aggregate.replay_target_sequence += item.replay_target_sequence;
         if (item.replay_catch_up_required) aggregate.replay_catch_up_required = true;
+        if (item.enrichment_failed) aggregate.enrichment_failed = true;
         aggregate.catch_up_applied_sequence += item.catch_up_applied_sequence;
         aggregate.catch_up_target_sequence += item.catch_up_target_sequence;
         if (item.catch_up_active) aggregate.catch_up_active = true;
@@ -1035,10 +1037,11 @@ fn embeddingsArtifactVisible(item: anytype, sparse: bool) bool {
     return item.doc_count > 0 and (item.node_count > 0 or item.root_node > 0);
 }
 
-fn backfillState(index_type: ApiIndexType, active: bool, replay_applied_sequence: u64, replay_target_sequence: u64, enrichment: ?db_mod.types.EnrichmentStats) []const u8 {
+fn backfillState(index_type: ApiIndexType, active: bool, enrichment_failed: bool, replay_applied_sequence: u64, replay_target_sequence: u64, enrichment: ?db_mod.types.EnrichmentStats) []const u8 {
     if (index_type == .embeddings) {
         _ = replay_applied_sequence;
         _ = replay_target_sequence;
+        if (enrichment_failed) return "failed";
         if (active) {
             if (enrichment) |stats| {
                 if (stats.worker_failed) return "failed";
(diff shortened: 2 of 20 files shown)


markhayden:
## Independent verification — built `a8ef15bd` clean, ran the full issue matrix against a live swarm

Thanks for the fast turnaround on this batch. We (downstream, Bakin) filed #201–#208 from a live integration smoke and re-verified this PR head directly: built `fix/issues-201-208` @ `a8ef15bd` from a clean tree with **no local patches** (zig 0.16.0, `zig build install`), ran `antfly swarm` against shared models, and probed `/ai/v1/embed` + `/db/v1` over loopback. Environment: darwin-arm64, single-node swarm, auth off.

**Every result below was reproduced on two independent labs** (separate data-dir *and* separate models-dir, the second with models **freshly re-pulled by this binary** to rule out stale-artifact false-negatives) — the embedding cosines matched to 4 decimals across both, so these are deterministic, not flaky.

### Scorecard

| Issue | Verdict | Evidence |
|---|---|---|
| **#201** media embedding + UAF | ✅ **fixed** | clipclap `remoteMedia` enrichment over loopback reaches `backfill_state: ready, total_indexed: 1` end-to-end. Both the `InvalidUri`/`UnsupportedEmbeddingProvider` gate and the binary-part corruption are gone. |
| **#204** `layoutStatsWithInvertedDetails` panic | ✅ **fixed** | 9-table create + immediate 20-doc batch burst while hammering `/indexes` status reads — **no panic, server stayed up**. `term_count` now populates (17) where it read 0 before. |
| **#202** split-CLIP `InputArityMismatch` | ⚠️ **partial** | The filed bug is fixed — `textInputTensorSet` graph-probing replaces the hardcoded `[input_ids, attention_mask]`, so the arity error is gone. But `Xenova/clip-vit-base-patch32` text embed now advances to **`500 ShapeMismatch`** (next layer down — fixed-length CLIP context vs dynamic padding). Still cannot text-embed. |
| **#207** sibling-index blast radius | 🟡 **likely fixed, unverifiable** | The isolation code is present (`enrichment_runtime.zig`), but create-time validation now rejects every failure trigger we tried (bad model → 500 at create; bad shape → 500 at create), so we can't reproduce a *runtime* single-index failure to confirm siblings survive. Needs a maintainer test that fails exactly one index of a multi-index table at enrichment time. |
| **#203** mxbai reranker | ❌ **not fixed** | Byte-identical `executeNode failed node_id=1091 op=transpose shape={0,0,0} err=error.UnsupportedShape` → query 500. Server stays alive (the #194 no-panic work holds), but reranking is still non-functional on both backends. |
| **#205** clipclap text degeneracy | ❌ **not fixed** | `/ai/v1/embed` text vectors still collapse: `cat`~`kitten` 0.984, `cat`~`"quarterly financial report"` 0.983; cross-modal scores sit in a 0.19–0.23 noise band (a spreadsheet query out-scores `"a photo of a cat"` against the actual cat photo). The CLIP **EOT-pooling fix you added is in `architectures/clip.zig`** (the ONNX path) — but clipclap loads on **metal/GGUF** and its text routes through `session_factory.zig`'s `.clip`/`.clap` dispatch, which the pooling change doesn't reach. Our standing hypothesis (unconfirmed): plain text is routing to the **CLAP** (audio-text) tower rather than CLIP, which would produce exactly this signature. The image side embeds fine (enrichment ready), so it's specifically the text encoder. |
| **#206** `_index_scores` dropped | ❌ **not fixed** | Fused hits still serialize `_index_scores: null`; `search/fusion.zig` computes per-index scores but the response path drops them. (This is the telemetry gap that masked #205 from us originally — per-leg debug surfaces read 0.0000.) |
| **#208** GGUF under `onnx` pin | ⏸️ **not retested** | Our labs ran auto-backend (clipclap → metal). The `TERMITE_PREFERRED_BACKEND=onnx` hard-fail-with-`NoModelFileFound` path wasn't exercised this round. |

### Net for downstream

The plumbing fixes (#201, #204) are real and we're glad to have them. But the two issues that actually gate **text→image retrieval quality — #205 and #202's downstream `ShapeMismatch` — both remain**, so multimodal search still returns noise on this head. #203 and #206 are also still reproducing as filed.

### Highest-leverage next step (#205)

If text embedding for clipclap is dispatching to the CLAP text encoder instead of CLIP in `session_factory.zig`, that single routing fix likely closes #205. A fast pass/fail probe for it (no corpus needed): embed `"a photo of a cat"`, `"a small kitten in a window"`, and `"quarterly financial report"` via `/ai/v1/embed` against clipclap — a healthy CLIP text tower gives cat~kitten ≫ cat~finance; the current build gives ~0.98 for both. (BGE through the identical harness discriminates correctly, so it isn't the test.)

Happy to re-run the full matrix on any follow-up commit, and to provide the repro scripts / cosine matrices for #205 if useful.

PR #215, opened 2026-06-09
Restore Go portable AFB backups in Zig

dovinmu:
> ## Problem
> 
> The local kind cloud-dev flow couldn't bring up Antfly Inference on Apple Silicon. Three image/config pins in
> the local kind path had drifted out of sync with the Termite → Antfly Inference rename (and with what `infra/`
> deploys to dev):
> 
> - The operator image (`DEV_OPERATOR_IMG`) ran legacy **Termite** controllers and crash-looped trying to watch
> CRDs the renamed, inference-only RBAC no longer grants — so the `InferencePool` was never reconciled and the
> StatefulSet was never created.
> - The runtime image (`DEV_ANTFLY_RUNTIME_IMG`) predated the rename (only had the old `termite` subcommand, not
> `inference`) and was effectively amd64-only.
> - `kind-deploy-operator` never passed the operator's `--inference-antfly-image` flag, so it fell back to the
> amd64-only `ghcr.io/antflydb/antfly:omni` default and inference pods `ImagePullBackOff`'d on arm64.
> 
> Dev (GKE) was immune because it pins the operator independently in `infra/` (`v0.0.17-rc.2`) and runs on amd64
> nodes — so "it works on dev / from main" didn't transfer, and merging main didn't help (the stale pins are on
> main too).
> 
> ## Changes
> 
> **Fix** (`Makefile`)
> - `DEV_OPERATOR_IMG` → `antfly-operator:v0.0.17-rc.2` (post-rename, matches infra dev)
> - `DEV_ANTFLY_RUNTIME_IMG` → `antfly:v0.2.0-rc.2` (post-rename, multi-arch amd64+arm64)
> - `kind-deploy-operator` now passes `--inference-antfly-image=$(ANTFLY_ANTFLY_INFERENCE_LOCAL_DEV_IMG)` so the
> operator provisions pods with the local multi-arch image
> 
> **Harden**
> - `devtools/kind/antfly-inference-diagnose.sh` (new) — read-only diagnostic that turns the opaque "Timed out
> waiting for StatefulSet" into specific drift warnings (pre-rename operator vs renamed RBAC, amd64-only runtime
> image on arm64, runtime image missing the `inference` subcommand), each pointing at the Makefile var to fix
> - `kind-deploy-antfly-inference-sample` runs the diagnostic on both failure paths; adds a
> `kind-antfly-inference-diagnose` target
> - `local-dev-doctor.sh` calls the diagnostic, and its operator-image probe is fixed — the old `{range
> …args[*]}` jsonpath always returned empty, so it had silently reported the operator image as `unset`
> - `local-cloud-tmux.sh` pins `HEALTH_PORT` per process (swarm/cloudaf=4202, jobsaf=4203) so they stop
> colliding with auth's default 4200, which had been failing the tmux pane health check (and making `cloudaf up`
> report a false failure)
> 
> ## Testing
> 
> Verified end-to-end on a local kind cluster (Apple Silicon):
> - `InferencePool` reaches `Running` (1/1), StatefulSet on `v0.2.0-rc.2`
> - `POST /ai/v1/embed` returns a 512-dim embedding
> - `make kind-deploy-antfly-inference-sample` exits 0
> - diagnostics: 0 warnings on a healthy stack; reversible bad-state test (flip operator to the omni default)
> fires the arch warning, then clean again
> - `cloud-ui` tmux session: auth on 4200, swarm health on 4202, no collision, dashboard serves on :3000


ajroetker:
```
  1. zig/pkg/antfly/src/api/http_server.zig:5527

     Cluster restore in swarm mode can report success without restoring any data. LocalSwarmMetadata.restoreTable only writes table/range metadata (zig/pkg/antfly/src/swarm/runtime.zig:314), and single-table restore compensates by calling restoreLocalTableDataFromManifest when self.cfg.swarm_mode is true (http_server.zig:5185). The cluster restore path does not do that; it marks the table "triggered" and continues at http_server.zig:5546. Result: POST /db/v1/restore in local swarm can create the restored tables but leave them empty. Mirror the single-table swarm data-apply path before marking the table triggered.

  2. zig/pkg/antfly/src/raft/storage/backup_restore.zig:202

     Remote portable restore idempotency/restart is broken. restoreSnapshotDocCount stages the snapshot with stageRestoreSnapshot before checking .afb, and remote stageRestoreSnapshot downloads a directory via copyDirectoryFromLocation. For portable backups the snapshot is a single AFB object, so a restart/reconcile that hits restoreSnapshotMatchesPath will try to stage a file as a directory before calling portableSnapshotDocCount. Check the .afb suffix first and use the file staging path (stageRestoreFile) for portable snapshots.
```

ajroetker:
```  
1. Medium: --format portable for single-table backup is accepted but ignored.
     .worktrees/pr-215-portable-afb-restore/zig/pkg/antfly/src/cmd/cli/backup.zig:55 now sends .format = format, and the OpenAPI request type includes format, but .worktrees/pr-215-portable-afb-restore/zig/pkg/antfly/src/api/
     public_table_http.zig:393 dispatches only backup_id and location. The TableApi vtable has no format parameter, and .worktrees/pr-215-portable-afb-restore/zig/pkg/antfly/src/api/http_server.zig:4218 always calls the existing physical
     snapshot backup path. So antfly backup --table docs --format portable ... can return success while writing a native backup, which is pretty misleading for this PR.
```

PR #217, opened 2026-06-10
Zig query perf: live-doc visibility, search effort policy, stale payload repair

dovinmu:
> Three query-speed fixes for the Zig runtime, found while chasing down the antfly-circus nightly numbers. Each commit is independently revertible and carries its own measurements; everything below is from the 50K OpenAI vdbbench case (k=100, serial HTTP, M4 Max) and big-ann random-xs.
> 
> ## 1. Broad live-doc visibility derivation (`e9316d8`)
> 
> A single tombstone — or any doc created after the read generation — made every dense query materialize the full live-doc set: a primary-namespace scan plus **two LSM point lookups per document**, and then a second per-doc liveness probe when the derived include set was re-filtered through the live filter it had just come from. ~1.8s/query at 50K docs, scaling linearly with table size (the old benchmark client's insert+delete healthcheck probe tripped this on every nightly run).
> 
> - `doc_identity`: three bulk range scans joined through arena hash maps, same contracts (incomplete-identity fallback, per-generation visibility).
> - `db`: memoize the resolved set per identity read generation; trust the visibility summary verdict in both directions.
> - `search_exec`: don't re-apply the live filter to its own output.
> 
> **Tables with 1 or 500 tombstones: ~1.8s → ~5ms per query, recall unchanged (0.983).**
> 
> ## 2. Dense search effort policy (`c2df4ed`)
> 
> - Width floor was `max(k, 64)` *leaves* — a k=100 query scanned ≥100 leaves even at effort 0, and on a ~100-leaf index the effort knob was inert. Now `ceil(4k/leaf_size)`.
> - ε resolved to 7–100, so the `(1+ε)×best` leaf pruning could never fire. Now 0.9–2.0.
> - Rerank retention was `ceil(ε+1)×k` = 8×k at default and **101×k** at effort 1.0; retention saturates ~8×k. Now a decoupled `SearchRequest.rerank_factor` (6–12×k).
> - Width ramps geometrically through a calibrated balanced anchor (70% of the leaf ceiling, absolute cap 2048 leaves so the default budget stops growing linearly with corpus size). Constants are env-overridable (`ANTFLY_SEARCH_*`) for offline tuning.
> 
> **50K defaults: p99 9.1ms → 3.5ms, avg 4.2ms → 3.1ms, recall 0.981 → 0.983. big-ann random-xs: +16% QPS at iso-recall 0.984, +41% at effort 1.0, effort now spans recall 0.87–0.99.**
> 
> ## 3. Stale dense posting payload repair (`c6e83d5`)
> 
> Bulk loads leave ~23% of visited leaf postings with `payload_dirty` set (deferred bulk splits), and nothing in the swarm serving path ever ran posting maintenance — so those leaves fell back to exact member scoring **on every query, forever** (~4.7K raw vector loads + full-precision distances per query at 50K). The background LSM maintenance worker now drains a bounded posting-maintenance round per wake (leases cached DBs, skips active bulk sessions, default 64-postings budget). One tick after a load, queries report zero stale-payload leaves.
> 
> Also adds `leaf_payload_stale`/`leaf_payload_missing` to the search profile and dense bench log line, which is how this was found.
> 
> ## Testing
> 
> - `zig build db-test` (1918 tests), `lib-vectorindex-test`, `provisioned-query-visibility-test` — green on the final state and at each commit boundary's subsystem.
> - Policy unit tests rewritten for the new semantics (floor, geometric ramp + cap, ε range, rerank factors).
> - Bench validation included tombstone-bearing permutations (1 and 500 tombstones) specifically to guard the visibility fast path, plus effort sweeps for the recall/latency frontier.
> 
> Known follow-ups (not in this PR, documented in #devlogs > fable-antfly-yolo): index build quality (identical loads vary ±0.015 full-scan recall; the tree needs ~⅔ of its leaves for 0.98), SIMD for the per-leaf query quantization, per-request serving overhead (the ~2.5K QPS ceiling on tiny datasets).
>

### DIFF
--- a/zig/build.zig
+++ b/zig/build.zig
@@ -2807,6 +2807,7 @@ pub fn build(b: *std.Build) void {
         "data runtime structural changes preserve writer-published runtime status",
         "data runtime startup catch-up prefers cached admin snapshot",
         "data runtime provisioned root refresh spawn failure preserves retry bookkeeping",
+        "data runtime background maintenance is due for dense posting cadence without lsm debt",
         "data runtime local split fallback preserves source identity namespace",
         "data runtime local merge fallback derives receiver identity namespace from catalog",
         "data public API listener uses public API request body limit",
--- a/zig/lib/vectorindex/src/hbc_index.zig
+++ b/zig/lib/vectorindex/src/hbc_index.zig
@@ -2101,6 +2101,14 @@ fn scoreLeafMemberIds(
         }
     }
 
+    if (self.config.use_quantization) {
+        if (!leaf_has_fresh_stored_payload) {
+            profile.leaf_payload_stale += 1;
+        } else {
+            profile.leaf_payload_missing += 1;
+        }
+    }
+
     const fetch_member_ids = scratch.member_ids[0..member_ids.len];
     var fetch_count: usize = 0;
     for (member_ids) |member_id| {
--- a/zig/lib/vectorindex/src/hbc_index.zig
+++ b/zig/lib/vectorindex/src/hbc_index.zig
@@ -1704,7 +1704,7 @@ pub fn searchProfiledRequest(
     };
     const search_width = req.search_width orelse self.config.search_width;
     const epsilon = req.epsilon orelse self.config.epsilon;
-    const rerank_factor: usize = search_mod.rerankFactor(epsilon);
+    const rerank_factor: usize = req.rerank_factor orelse search_mod.rerankFactor(epsilon);
     const should_rerank = self.config.use_quantization and self.config.rerank_policy != .never;
     const candidate_limit: usize = if (should_rerank) req.k * rerank_factor else req.k;
     const candidate_capacity: usize = search_mod.candidateCapacity(search_width, self.metadata.branching_factor);
--- a/zig/lib/vectorindex/src/search_types.zig
+++ b/zig/lib/vectorindex/src/search_types.zig
@@ -75,6 +75,10 @@ pub const SearchProfile = struct {
     approx_leaves_scored: u64 = 0,
     approx_vectors_scored: u64 = 0,
     exact_vectors_scored: u64 = 0,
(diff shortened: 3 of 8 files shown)


ajroetker:
```  
- P1: Dense posting maintenance can be stranded when there is no LSM maintenance debt. The worker still sleeps/skips when lsmMaintenanceScoreBestEffort() == 0 in .worktrees/pr-217-review/zig/pkg/antfly/src/data/runtime.zig:2249, and startup
    also refuses to wake/spawn the worker on zero LSM score in .worktrees/pr-217-review/zig/pkg/antfly/src/data/runtime.zig:2215. The new dense posting repair only runs later in that worker at .worktrees/pr-217-review/zig/pkg/antfly/src/data/
    runtime.zig:2302. If a bulk load leaves dirty dense posting payloads but no LSM debt, the repair will not run until unrelated LSM work or an explicit wake happens. The fix should make dense posting backlog part of the wake/eligibility check,
    or give it a separate maintenance wake path.
```

dovinmu:
> ```
> - P1: Dense posting maintenance can be stranded when there is no LSM maintenance debt. The worker still sleeps/skips when lsmMaintenanceScoreBestEffort() == 0 in .worktrees/pr-217-review/zig/pkg/antfly/src/data/runtime.zig:2249, and startup
>     also refuses to wake/spawn the worker on zero LSM score in .worktrees/pr-217-review/zig/pkg/antfly/src/data/runtime.zig:2215. The new dense posting repair only runs later in that worker at .worktrees/pr-217-review/zig/pkg/antfly/src/data/
>     runtime.zig:2302. If a bulk load leaves dirty dense posting payloads but no LSM debt, the repair will not run until unrelated LSM work or an explicit wake happens. The fix should make dense posting backlog part of the wake/eligibility check,
>     or give it a separate maintenance wake path.
> ```

Okay, this should be fixed. We should be good to merge!

PR #224, opened 2026-06-11
Fix raft apply deadlock behind nightly e2e hangs; harden CI diagnostics

ajroetker:
Fixes the two failing nightly jobs from run 27327168835 (`e2e-base`, `full-default`), which turn out to share a root cause. Diagnosed by reproducing the hang in a Linux container and taking gdb thread dumps mid-wedge.

## The deadlock (e2e-base: `test_multinode_autograph_resolves_promotes_and_hydrates_entities`)

Mid-hang stacks show the **raft apply thread** wedged:

```
MultiRaft.runRound → flushPendingApply → applyReady
  → applyReplicatedBatchGroupLocal → getOrOpenCachedDbMode
  → drainResolverBackfill → runUntilIdle → PromotionRuntime.catchUp → lockMutex   (blocked ≥30s)
```

`getOrOpenCachedDbModeAtGeneration` unconditionally ran `db.drainResolverBackfill()` after every cache-layer open — including `.default_async` opens on the raft apply thread (the inner `DB.open` even passes `drain_resolver_backfill = false`, but the cache layer drained anyway). The drain runs promotion catch-up, whose cross-shard entity upserts need raft applies that are queued behind this very open. Result: the raft loop blocks, every group's applies stall, batch writes time out cluster-wide, and nothing is logged (pure in-memory waits). This matches all three observed e2e failure modes (doc:b insert timeout, create_table timeout, "entities were not promoted").

**Fix:** skip the synchronous drain for `.default_async` opens. The promotion/resolution workers started by the open drain the same backlog asynchronously.

## Spin-lock starvation (full-default: "runner lost communication")

Two helpers spin with no yield (`while (!tryLock()) spinLoopHint()`): `data.runtime.lockAtomic` and `api.table_writes.lockAtomic`. The dumps show waiters (e.g. `waitForLocalRaftBatchApply` behind the wedged raft loop) pinning a core each. On CPU-quota'd ARC pods this starves the lock holder — and the runner agent, which stops heartbeating; GitHub declares the runner lost after ~10 min, which is how most `full-default` failures present (death consistently ~1s into an in-process cluster sim test, runner silent exactly 10 minutes before the job dies). Both now bounded-spin then `Thread.yield()`, with the freestanding comptime guard used elsewhere (`clock.zig`). ~28 more pure-spin sites exist across the tree — worth a shared helper as follow-up.

Additional `full-default` hardening: heavy zig steps run under `taskset -c 0-7`, bounding both `zig build` parallelism and `getCpuCount()`-derived thread pools inside the cluster sims (sized from node-visible cores, not pod quota).

## CI diagnosability (how this stays fixed)

- `MultiNodeScalingCluster.native_stack_dumps()` — gdb `thread apply all bt` for every node process, attached to batch-insert and promotion-timeout failures in `test_resolution` (gdb added to workflow deps). The next silent wedge arrives with stacks in the CI failure output; this is how this PR's deadlock was found.
- `ANTFLY_FS_PATH_DEBUG` no longer defaults on in the e2e script: ~8k lines/s/node of path-component logging slowed servers and made the captured failure tails pure noise (every prior CI dump of this hang was undiagnosable because of it). Failure log tails raised 20KB → 200KB.

## Validation

- Container repro (4 CPUs, Linux, Debug, CI-like env): failed within 2 iterations before the fix, both failure modes; **19/20 after** even with fs-path logging still enabled, and the remaining failure's stack dumps show no lock wedge (pure resource saturation at 4 CPUs: all nodes mid-compaction/encoding). One residual finding for follow-up: `status_only` opens hold `local_db_mutex` across a full DB open, queueing lookups behind status refresh on slow machines.
- Native: `zig build integration-test` 55/55; lifecycle + resolution e2e green.
- The wasm target fails to build on main for pre-existing reasons (`mem_backend.zig` `Thread.yield`, 64-bit atomics in `promotion_runtime.zig`) — untouched by this PR.

### DIFF
--- a/.github/workflows/zig-tests.yml
+++ b/.github/workflows/zig-tests.yml
@@ -150,7 +150,7 @@ jobs:
           submodules: true
 
       - name: Install build tools
-        run: sudo apt-get update && sudo apt-get install -y build-essential pkg-config ca-certificates git curl xz-utils
+        run: sudo apt-get update && sudo apt-get install -y build-essential pkg-config ca-certificates git curl xz-utils gdb
 
       - name: Install uv
         uses: astral-sh/setup-uv@v8.1.0
@@ -191,7 +191,7 @@ jobs:
           submodules: ${{ needs.changes.outputs.trace_validate == 'true' }}
 
       - name: Install build tools
-        run: sudo apt-get update && sudo apt-get install -y build-essential pkg-config ca-certificates git curl xz-utils
+        run: sudo apt-get update && sudo apt-get install -y build-essential pkg-config ca-certificates git curl xz-utils gdb
 
       - name: Set up Zig
         if: needs.changes.outputs.trace_validate == 'true'
@@ -262,7 +262,7 @@ jobs:
       - uses: actions/checkout@v6
 
       - name: Install build tools
-        run: sudo apt-get update && sudo apt-get install -y build-essential pkg-config ca-certificates git curl xz-utils
+        run: sudo apt-get update && sudo apt-get install -y build-essential pkg-config ca-certificates git curl xz-utils gdb
 
       - name: Install uv
         uses: astral-sh/setup-uv@v8.1.0
@@ -279,20 +279,33 @@ jobs:
       - name: Verify Zig version
         run: zig version
 
+      # The in-process cluster simulations size their thread pools from the
+      # visible CPU count, which on ARC reflects the node rather than the pod
+      # quota. The resulting burst can starve the runner agent long enough for
+      # GitHub to declare the runner lost (~10 min of missed heartbeats), which
+      # is how most full-default failures present. Bound the test steps to a
+      # fixed CPU set so pools and zig build parallelism stay within budget.
+      - name: Compute CPU bound for heavy test steps
+        id: cpus
+        run: |
+          jobs=$(nproc)
+          if [ "$jobs" -gt 8 ]; then jobs=8; fi
+          echo "list=0-$((jobs - 1))" >> "$GITHUB_OUTPUT"
+
       - name: Run full Zig unit tests
-        run: make zig-unit-test
+        run: taskset -c ${{ steps.cpus.outputs.list }} make zig-unit-test
 
       - name: Run Zig simulation tests
         working-directory: zig
-        run: zig build sim-test
+        run: taskset -c ${{ steps.cpus.outputs.list }} zig build sim-test
 
       - name: Run Zig integration tests
         working-directory: zig
-        run: zig build integration-test
+        run: taskset -c ${{ steps.cpus.outputs.list }} zig build integration-test
 
       - name: Run Zig recall tests
         working-directory: zig
-        run: zig build recall-test
+        run: taskset -c ${{ steps.cpus.outputs.list }} zig build recall-test
 
       - name: Run Zig recall harness sweep
         working-directory: zig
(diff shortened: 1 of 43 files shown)


ajroetker:
Follow-up commit `0350db0d4`: consolidated all ~33 try-lock spin loops into a shared `lib/platform/src/sync.zig` (exported as `platform.sync` on the `antfly_platform` module) rather than leaving per-file copies:

- `lockYielding(mutex)` — bounded spin (64), then `platform.time.yieldBriefly()` (the existing freestanding-guarded 100µs sleep), for plain OS threads.
- `lockYieldingIo(mutex, io)` — reschedules via `io.sleep(Io.Duration.zero, .awake)`, matching the `io_threaded_runtime` idiom, for code already holding an `Io`.

Bonus: the unguarded `Thread.yield()` loops in `mem_backend.zig`/`lmdb.zig` were among the pre-existing wasm build breaks (no `sched_yield` on freestanding); they now compile under the guard. `pkg/antfly/src/platform/clock.zig` keeps its local spin since `platform/` doubles as the standalone module shim and can't import outside its root.

Container validation with the consolidated form: lock waiters now show up in gdb parked in `nanosleep` via `yieldBriefly` instead of pinning cores; integration 55/55, lsm-backend 221/221, resolution e2e 17–18/20 at 4 CPUs (residual failures are the documented no-lock resource-saturation mode).

ajroetker:
Commit `68b7d2513` addresses both failures from run 27369813582:

**`unit`** — `db primary lsm maintenance step does not reclaim index obsolete paths` asserted `obsolete_paths_reclaimable == 1` on a single stats snapshot, but a queued path transiently counts as `pinned_by_readers` whenever a background reader holds the backend — on a loaded runner the snapshot can land in that window. Both maintenance-step tests now poll the precondition (`expectObsoletePathsReclaimable`). Locally: db-test 1948 passed / 0 failed.

**`e2e-base`** — `test_autoscaling_drains_data_node_and_replaces_placements` hit `ConnectionResetError(104)` on its seed batch with zero diagnostics (bare `requests.post`). Two listener paths produce client-visible RSTs and are fixed:
- When connection-thread slots are saturated, the listener **accepted and immediately closed** new connections — an RST to a client that already wrote its request. It now serves the connection inline (blocking the accept loop is natural backpressure). Regression test added: with one slot held by a slow request, a second request still gets a 200.
- `receiveHead() catch return` closed silently with unread request bytes; non-idle-close failures are now logged so resets carry evidence.

The seed batch also attaches node logs + native stack dumps on failure now. The reset didn't reproduce in 8 container iterations of the scaling file (48 test executions) — it appears to need full-suite load — so if it recurs in CI, the failure output will carry the diagnosis. Note the resolution e2e (the original deadlock) passed in this run: 98 passed / 1 failed, with this scaling flake the only remainder.

ajroetker:
Good catch on the inline-serve fallback — an idle client that never sends headers would park the accept thread in `receiveHead()`, and `stop()`'s wake only unblocks `accept()`, so shutdown could hang behind that read. Reworked in `9d12aa67d` to the suggested shape:

- The accept loop now **acquires a connection-thread slot before calling `accept()`**. Under saturation it stops accepting and connections queue in the kernel backlog — no accept-then-close RST, no client reads on the accept thread. The slot-wait loop polls `stopping`, so shutdown stays prompt; the slot is released on every exit path and transfers to the connection fiber on successful handoff.
- Reworked the saturation regression test for the queue-then-complete semantics (second request must not be reset while the slot is held, and completes 200 once it frees).
- Added the suggested shutdown test: only slot held by a slow request + a headerless connection queued → `stop()` must return (it does; the headerless connection is never accepted).

Validation: listener buckets green including both new tests, integration 55/55, and 3/3 container iterations of the scaling + resolution e2e files with the new accept loop.

PR #260, opened 2026-06-21
 Improve Gemma 4 E2B Metal inference reuse and fix unbounded RAG generation

timkaye11:
## Summary

  Speeds up Gemma 4 E2B Metal inference by reducing repeated setup work and adding a warm-server path
  for local CLI/server generation.

Not sure if we want to keep the warm-server path, but it makes testing and makes the  time-to-token faster for the first request. Normally we'd be 'warm' after the first request anyway, but `warm generate` moves it to startup 

  ## Changes

  - Add mmap/no-copy Metal residency for Q8_0 GGUF weights.
  - Thread mapped weight offsets through Metal Q8 kernels.
  - Reuse per-model Metal whole-model graph/runtime caches in server generation.
  - Add `antfly inference run --warm-generator ... --warm-generator-backend metal`.
  - Let `antfly inference generate` use a warm server via `--server` / `--require-server`.
  - Add default `max_tokens = 256` for generator configs to avoid unbounded RAG generation.
  - Add focused Gemma 4 E2B Metal benchmark coverage.

  ## Validation

  ```sh
  zig build
  zig build test -- --test-filter "metal partition executor eager multi op chain matches host"
  zig build test -- --test-filter "parseArgs accepts server URL" "server generate routes metal
  requests to whole model" "default server auto route is metal only" "default server strips local
  models dir prefix"
  zig build root-test -- --test-filter "retrieval agent supports generation step in phase 2"
  git diff --check main

  ## Smoke Test

  cd zig && ./zig-out/bin/antfly inference run \
    --host 127.0.0.1 \
    --port 8090 \
    --models-dir ~/.antfly/inference/models \
    --warm-generator ggml-org/gemma-4-e2b-it-gguf \
    --warm-generator-backend metal

  Then:

  cd zig && ./zig-out/bin/antfly inference generate \
    ggml-org/gemma-4-e2b-it-gguf \
    'Give a one sentence summary of Korean history.' \
    --backend metal \
    --max-tokens 32 \
    --require-server \
    --print-timing \
    --print-token-count \
    --print-finish-reason


zig/lib/generating/src/mod.zig
```diff
@@ -99,6 +99,8 @@ pub const Provider = enum {
     }
 };
 
+pub const default_max_tokens: i64 = 256;
```
ajroetker, line 105: Is this the same default as openai/ollama?

zig/pkg/antfly/src/swarm/runtime.zig
```diff
@@ -2818,6 +2848,8 @@ fn printUsage() void {
         \\  --inference-combined-budget-mb <n>    Embedded inference native generation combined budget override
         \\  --inference-kv-budget-mb <n>          Embedded inference native generation KV cache budget override
         \\  --inference-scratch-budget-mb <n>     Embedded inference native generation scratch budget override
+        \\  --warm-generator <name>                Preload and warm an embedded generator model before serving (repeatable)
```
ajroetker: This should probably be more general and in the inference config like antfly had with termite in the go version where you can specify what models you want to warm across all types, we should also make sure it actually integrates with the resource_manager budgeting

specs/openapi/inference/api.yaml
```diff
@@ -2768,6 +2768,35 @@ components:
...
+            - embedder
+            - reranker
+            - chunker
+            - classifier
+            - recognizer
+            - rewriter
+            - reader
+            - transcriber
+            - extractor
+        name:
```
ajroetker, line 2831: Will we also want the tags (for quantization) or metadata like if it needs the projection model loaded too (for speculative decoding?)

specs/openapi/inference/api.yaml
```diff
@@ -2886,18 +2915,29 @@ components:
...
+            generation request so native/Metal weights, KV setup, and kernels use
+            the same budgeted path as request-time generation. Other model kinds
+            use the best available warm path for that kind.
+          items:
+            $ref: '#/components/schemas/WarmModel'
+          example:
+            - kind: generator
+              name: gemma-e2b
+              backend: metal
         preload:
```
ajroetker: Should we just update preload to be the new form that you added for warm_models? You could also preload the ml models too right?

specs/openapi/inference/api.yaml
```diff
@@ -2768,6 +2768,46 @@ components:
         arguments:
           type: string
           description: Incremental arguments JSON string
+    WarmModelKind:
```
ajroetker: Should this just be ModelKind and shared with the existing ModelKind?

specs/openapi/inference/api.yaml
```diff
@@ -2768,6 +2768,46 @@ components:
...
+        - embedder
+        - reranker
+        - chunker
+        - classifier
+        - recognizer
+        - rewriter
+        - reader
+        - transcriber
+        - extractor
+    WarmModelBackend:
```
ajroetker: Should this backend type be shared with the existing one?

specs/openapi/inference/api.yaml
```diff
@@ -2768,6 +2768,46 @@ components:
...
+      type: string
+      description: Optional backend preference for a warm target.
+      enum:
+        - native
+        - onnx
+        - metal
+        - cuda
+        - wasm
+    WarmModel:
```
ajroetker: Should this just be ModelRef or Model?

specs/openapi/inference/api.yaml
```diff
@@ -2886,18 +2926,29 @@ components:
             Requests exceeding this timeout receive 504 Gateway Timeout.
           default: '0'
           example: 30s
+        warm_models:
```
ajroetker: Can the warm_models be combined with preload (let's just use preload since it's already there)

specs/openapi/inference/api.yaml
```diff
@@ -2888,16 +2925,17 @@ components:
...
+            request so native/Metal weights, KV setup, and kernels use the same
+            budgeted path as request-time generation. Other model kinds use the
+            best available warm path for that kind.
+          items:
+            $ref: '#/components/schemas/ModelRef'
           example:
-            - BAAI/bge-small-en-v1.5
-            - openai/clip-vit-base-patch32
+            - kind: generator
+              name: gemma-e2b
```
ajroetker: This still needs the <owner>/<repo> format added back in I think and probably a `format`, `quantization` too for gguf/onnx/safetensors and `q4_k/q8/fp16`?

PR #313, opened 2026-07-02
TLA+ overhaul

dovinmu:
> # TLA+ models for Antfly's coordination protocols
> 
> ## What this is
> 
> 37 small TLA+ models of the protocols where a broken ordering, crash-recovery, or visibility rule silently loses or corrupts data: HA failover/replication, shard split/merge, snapshot transfer, transactions, WAL/LSM/LMDB commit paths, replay/enrichment watermarks, document identity, node scale-down, table create/drop, index rebuilds, backups vs WAL retention, CDC, write batching, and publication steps (Lite, ML artifacts, generated OpenAPI packages).
> 
> Each model states rules that must always hold — "a promoted standby has every synchronously-acknowledged write", "a delete then a write to the same key ends with the write", "WAL is never deleted while an active standby that hasn't been told to re-seed needs it" — and the TLC model checker explores every reachable interleaving of steps, crashes, and restarts within the model's bounds.
> 
> ## Layout
> 
> Specs live in directories mirroring the code they model (`specs/tla/metadata/`, `raft/`, `storage/{ha,db,lsm,lmdb,lite}/`, `api/`, `ml/`). All checks for a model — the passing configuration, its deliberately-broken variants, heavy variants — live in one `<Model>.cfgs` file as named sections in plain TLC config syntax, with tier membership annotated on the section header. Adding a model or a check touches no build files.
> 
> ## Build integration
> 
> One script and four make targets:
> 
> - `scripts/tla-check.sh` — the runner. Discovers checks from the `.cfgs` files; subcommands for the full gate, tiers, single checks, the deliberate-bug suite, parse-only smoke, a hygiene audit, and NDJSON trace validation.
> - `make tla-check` — full gate (audit + parse + core + fast + all 116 deliberate-bug checks); `TIER=heavy|manual` for the big runs; `CHECK=<id>` for one check.
> - `make tla-trace TRACE=<family> TRACE_FILES=...`, `make tla-clean`, and main's existing `tla-tools`.
> 
> Net Makefile change vs main is negative (−57 lines); there are fewer TLA make targets than main had.
> 
> ## Why a passing check is meaningful
> 
> - Every safety rule has a deliberately-broken variant: a check section that switches on one specific bad behavior. The suite fails unless TLC catches it **and names that exact rule** — a crash, typo, or unrelated failure does not count. 116 such checks. (25 older ones still trip a bundle of rules rather than one; the audit reports them as open migration debt.)
> - The hygiene audit (part of the default gate) fails on: stale checker artifacts, bad behaviors defined in a model but never exercised by any check, and check sections that no tier would ever run.
> - `MODEL_COVERAGE.md` grades every model by its correspondence evidence — executed test, recorded trace, or code reading — so a hand-modeled sketch can't be mistaken for a verified statement about the implementation.
> 
> ## Also in the change
> 
> - Progress (liveness) properties on the highest-stakes models: failover completes, catch-up drains, drains reach a safe state.
> - Deliberately not modeled, with reasons and named follow-ups: distributed join leases (simulation harness), background-job gate races (Zig tests), merge rollback (no abort path exists in the implementation).
> - Zig changes are tests only, no production code: replay hint-lane co-writes on decode failure (executed, passing) and split-preserves-latest-write (passing).
> 
> ## Validation
> 
> Full gate green: parse, 4 core + 34 fast positives, all 116 deliberate-bug checks tripping their named rule, hygiene audit clean. Heavy tier green (largest model explores ~37M distinct states). All trace-fixture validators green (10/10 fixtures).
> 
> ## Limits
> 
> Every model runs and passes; what varies is the evidence that a model matches the implementation — some are anchored by executed tests or recorded traces, others only by code reading, and the coverage inventory says which. Exploration is bounded (typically 2–3 writes/nodes per model; the heavy tier pushes the riskiest models to tens of millions of states). A green run means the modeled rules hold and are breakable in principle — design evidence, not implementation proof.

### DIFF
--- a/zig/bench/run_tokenizer_experiments.sh
+++ b/zig/bench/run_tokenizer_experiments.sh
@@ -1,0 +1,68 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+if [[ $# -lt 2 || $# -gt 4 ]]; then
+  echo "usage: $0 <tokenizer.json> <corpus.txt> [repeat] [exact|hash]" >&2
+  exit 2
+fi
+
+resolve_path() {
+  local input=$1
+  local directory
+  directory=$(cd -- "$(dirname -- "${input}")" && pwd)
+  printf '%s/%s\n' "${directory}" "$(basename -- "${input}")"
+}
+
+tokenizer_json=$(resolve_path "$1")
+corpus=$(resolve_path "$2")
+repeat=${3:-1}
+validation=${4:-exact}
+script_dir=$(cd -- "$(dirname -- "${BASH_SOURCE[0]}")" && pwd)
+zig_dir=$(cd -- "${script_dir}/.." && pwd)
+
+run_benchmark() {
+  (
+    cd "${zig_dir}"
+    zig build -Doptimize=ReleaseFast bench-tokenizer -- \
+      "${tokenizer_json}" "${corpus}" --repeat "${repeat}" \
+      --validation "${validation}" --mmap-corpus --prefault-corpus "$@"
+  )
+}
+
+echo "experiment=stage_diagnostics"
+run_benchmark \
+  --warmup 1 --iterations 1 --threads 1 --internal-threads 1 \
+  --diagnostics --diagnostic-iterations 1
+
+echo "experiment=cache_profile"
+run_benchmark \
+  --warmup 1 --iterations 1 --threads 1 --internal-threads 1 \
+  --profile-bpe
+
+for cache_mb in 0 2 8 32 64 128 256; do
+  echo "experiment=cache_capacity cache_max_mb=${cache_mb}"
+  run_benchmark \
+    --warmup 2 --iterations 5 --threads 1 --internal-threads 16 \
+    --cache-max-mb "${cache_mb}"
+done
+
+for bulk_slots in 4096 16384; do
+  echo "experiment=bulk_cache bulk_slots_per_shard=${bulk_slots}"
+  run_benchmark \
+    --warmup 2 --iterations 5 --threads 1 --internal-threads 16 \
+    --cache-max-mb 128 --cache-bulk-slots "${bulk_slots}"
+done
+
+for tasks in 1 2 4 8 12 16; do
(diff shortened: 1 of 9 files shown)


zig/specs/tla/api/AntflyOpenApiCodegen.tla
```diff
@@ -0,0 +1,334 @@
...
+    committedImports,
+    committedComplete,
+    checkPassed
+
+vars == <<sourceVersion, joinedPublicVersion, prefixedPublicVersion,
+          rootOpenApiVersion, generatedVersion, generatedMode,
+          generatedImports, generatedComplete, committedVersion,
+          committedMode, committedImports, committedComplete, checkPassed>>
+
+NoVersionByPackage == [p \in Packages |-> 0]
```
ajroetker: Does this make sense to have a TLA spec for?

zig/specs/tla/TLA_CRITIQUE_REPAIR.md
```diff
@@ -0,0 +1,110 @@
+# TLA+ Critique Repair Tracker
```
ajroetker: Implementation doc

zig/specs/tla/MODELING_PLAN.md
```diff
@@ -0,0 +1,1212 @@
+# Antfly TLA+ Modeling Plan
```
ajroetker: plan doc

zig/specs/tla/INVENTORY.md
```diff
@@ -0,0 +1,1106 @@
+# TLA+ Model Coverage Inventory
```
ajroetker: Should this be INVENTORY.md

zig/specs/tla/api/AntflyOpenApiCodegen.tla
dovinmu: Probably not; removed

ajroetker:
looks like merge conflicts!

PR #314, opened 2026-07-02
Fix provisioned write coalescer losing later writes to deleted keys

dovinmu:
> Independent batches coalesced into one provisioned group batch were merged by bucketing all writes and all deletes into separate arrays, and DB batch application resolves writes before deletes regardless of request order. A delete(k) enqueued before a write(k, v) in the same coalesced flush therefore ended with k deleted and the acknowledged write silently lost.
> 
> Merge now stages operations per key in waiter enqueue order (last operation wins, mirroring the BulkIngestCoalescer staging pattern), so the merged write/delete key sets are disjoint and downstream ordering cannot invert.
> 
> The new correspondence test, motivated by the AntflyBatcherCoalescing TLA+ model's LastOperationWinsPerKey invariant, fails without this fix and passes with it. The existing same-group coalescing test now also covers write-then-delete on one key.


ajroetker:
```• Findings

  - High: /private/tmp/antfly-pr314-review/zig/pkg/antfly/src/api/table_writes.zig:6921 can now acknowledge an invalid
    waiter as successful when its operation is overwritten by a later coalesced waiter on the same key.
    stageCoalescedEntriesLastOp drops the earlier op before applyProvisionedGroupBatchLocked validates the merged batch,
    then finishCoalescedEntries(entries, null) marks every waiter successful. I reproduced this with a temporary local
    test: invalid JSON write to doc:order, followed by delete of doc:order; the invalid worker’s err stayed null. This
    weakens the existing failure isolation contract. A conservative fix is to avoid coalescing overlapping-key waiters, or
    fall back to individual application whenever staging would drop any original op.

  Verification

  - Clean PR target passed before the scratch repro: env ZIG_GLOBAL_CACHE_DIR=/private/tmp/antfly-pr314-zig-global-cache
    zig build api-table-writes-docid-test from zig/ produced 30 passed; 0 failed.

  - Scratch regression failed as expected with invalid.err == null.
  - GitHub e2e-base is failing, but the pulled job log shows a ReadTimeout in e2e/antfly/
    test_resolution.py::test_multinode_autograph_resolves_promotes_and_hydrates_entities while creating the documents
    table; I did not see evidence tying that CI failure directly to this patch.```

**Comment by dovinmu** [comment]
```
High finding fixed: overlapping-key waiters fall back to individual application, staging machinery deleted, regression test added; api-table-writes-docid-test 31/31.

The failure is not caused by PR 314. The evidence is structural and decisive, not a judgment call:
- test_multinode_autograph_... runs over a live multi-node raft cluster (its own docstring says so). PR 314 changes the provisioned/non-raft write coalescer — it is physically not in this test's execution path.
- AJ independently saw the same ReadTimeout in this exact test during his review and explicitly declined to attribute it to the patch.
- e2e-base has been flaking on different tests across unrelated branches (backup-restore on your TLA branch, graph traversal on AJ's http-listener branch, resolution here) — the signature of an
overloaded/timing-sensitive e2e job, not a per-PR regression.
```

dovinmu:
Maybe we should keep this open and see if this is a flaky test issue

PR #324, opened 2026-07-03
Handle unsupported OCR backend paths

stinkbugaf:
## Summary
- reject unsupported strided ConvTranspose imports in native ONNX so incompatible OCR detector graphs fall through instead of producing invalid output
- retry multi-stage OCR reader loading per backend candidate so the whole pipeline lands on a compatible backend
- tighten dynamic shape handling for ONNX Sigmoid, Conv, Resize, and runtime reshape paths, with focused regression coverage
- add generic detector-output tracing behind ANTFLY_INFERENCE_TRACE_OCR_DETECTION for future OCR diagnostics

## Why
Some OCR reader models can be partially accepted by the native backend even when required graph semantics are not implemented exactly. That can produce malformed detector outputs and empty OCR results. This change makes unsupported native paths explicit and lets backend selection recover model-agnostically.

## Verification
- zig build install -Dedition=inference -Dmetal=false -Donnx=true
- native-preferred multi-stage OCR e2e passes by falling back to a compatible backend
- focused ONNX/import/runtime regression tests pass
- git diff --check

### DIFF
--- a/zig/build.zig
+++ b/zig/build.zig
@@ -1913,6 +1913,7 @@ pub fn build(b: *std.Build) void {
         .optimize = .ReleaseSafe,
         .single_threaded = true,
     });
+    wasm_inference_ml_mod.addImport("antfly_platform", wasm_platform_mod);
     const wasm_inference_onnx_graph_mod = b.createModule(.{
         .root_source_file = b.path("lib/onnx/src/root.zig"),
         .target = wasm_target,
--- a/zig/build.zig
+++ b/zig/build.zig
@@ -2214,6 +2214,13 @@ pub fn build(b: *std.Build) void {
     const lib_ml_tabular_test_step = b.step("lib-ml-tabular-test", "Run standalone lib/ml/tabular tests");
     lib_ml_tabular_test_step.dependOn(&run_lib_ml_tabular_tests.step);
 
+    const lib_onnx_tests = b.addTest(.{
+        .root_module = termite_onnx_graph_mod,
+    });
+    const run_lib_onnx_tests = b.addRunArtifact(lib_onnx_tests);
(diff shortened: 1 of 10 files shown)


ajroetker:
Real issue 
```
• This looks like a real deterministic build issue, not a flaky test.

  The failing job unit on run 28687288705 fails while compiling the newly added standalone lib/onnx test target:

  lib/std/c.zig:10719:12: error: dependency on libc must be explicitly specified in the build command
  referenced by:
      fuseSDPA: lib/ml/src/graph/passes/fuse.zig:2800:14

  PR #324 adds lib_onnx_tests in zig/build.zig and wires it into unit-test. That ONNX test imports lib/ml, and fuseSDPA calls std.c.getenv("ANTFLY_DISABLE_SDPA_FUSION"). The generated zig test command does not include -lc, so Zig rejects the
  std.c dependency.

  I reproduced the same compile error locally with the same command shape. Adding -lc made the compile pass with --test-no-exec, so the cause is the missing libc linkage, not timing or test nondeterminism.

  Likely fixes: either link libc for the new lib_onnx_tests/related modules, or better, guard the std.c.getenv call behind builtin.link_libc so no-libc test builds do not depend on libc just for an optional env var.
```
`std.c.getenv` in those places needs to be replaced with the lib/platform version

stinkbugaf:
Addressed in cc5d30f7b.

Changed the SDPA fusion env guard in lib/ml to use antfly_platform.env instead of std.c.getenv, and wired antfly_platform into the lib/ml module. Verified the standalone ONNX test target now compiles without requiring libc:

- zig build lib-onnx-test -- --test-no-exec
- zig build lib-onnx-test -- --test-filter "ConvTranspose 2D with stride"
- zig build lib-onnx-test -- --test-filter "Sigmoid preserves dynamic input shape"
- git diff --check

PR #330, opened 2026-07-06
Antfly Lite: cache native pages, chain links, and docstore snapshots

dovinmu:
> ## What
> 
> Three storage-layer caches for the Antfly Lite native backend, all semantics-preserving (no on-disk format, API, or durability changes). Together they remove the per-transaction whole-DB re-materialization that dominated Lite read/write latency.
> 
> 1. **Page cache in `NativeFile`** — raw encoded pages cached by page id (64 MiB cap). Updated in `writePage`, cleared on vacuum/file-replacement. Safe because pages are immutable for the life of an open handle: in-process writes flow through `writePage` (updates cache) or `replaceOpenFileWithImage` (clears it), and free-page reuse/vacuum require the exclusive data-rewrite lock that live handles block. `check()` holds a bypass so **integrity checks read disk, not cache**, and free-map pages are never cached so free-list validation stays on disk truth.
> 2. **Link cache** — decoded chain-navigation metadata (`previous_page`/`next_page`, external-value refs, chunk lengths, catalog keys) cached per page, so per-commit free-map reachability walks traverse chains as hash lookups instead of pread+decode per page. Same walk algorithm, byte-identical results; separately budgeted (16 MiB).
> 3. **Refcounted docstore snapshot** — the materialized sorted doc array is cached on the `Store` keyed by `(commit_sequence, document_root_page)`; transactions retain/release it instead of each re-walking the whole chain. Commits apply their own mutations to produce the next snapshot in memory. Pinned-snapshot semantics, read-your-writes, and cross-process behavior are unchanged; vacuum / out-of-band catalog commits invalidate via the key.
> 
> ## Why
> 
> The Lite native docstore re-walked and re-materialized the entire document chain per transaction, giving a flat, size-independent per-op floor (µs–ms on Apple Silicon, tens–hundreds of ms on older x64). This is what made embedded/CLI consumers (e.g. bit-refinery) storage-bound.
> 
> ## Performance
> 
> Benchmarked before/after via the Go binding (`go/pkg/antflylite`), ReleaseFast, 209 docs × ~2.3 KB ("bit-refinery shape").
> 
> **Apple Silicon (dev machine):**
> 
> | op | before | after | Δ |
> |---|---|---|---|
> | point lookup | 1.60 ms | 1.4 µs | ~1,150× |
> | point lookup after churn | 19.7 ms | 0.75 µs | ~26,000× |
> | full scan (209) | 1.97 ms | 0.85 ms | ~2.3× |
> | 1-doc durable write | 12.9 ms | 0.87 ms | ~15× |
> | mixed 1 write + 10 reads | 35.1 ms | 0.79 ms | ~44× |
> | churn 2000 updates | 96.5 ms/op | 5.9 ms/op | ~16× |
> 
> **Independent cross-hardware verification (Linux x64, i7-7700HQ 2016, LUKS SSD):** same suite, before = shipped rc.17 (uncached), after = this branch built locally.
> 
> | op | before | after | Δ |
> |---|---|---|---|
> | point lookup | 4.28 ms | 3.3 µs | ~1,300× |
> | point lookup after churn | 82.1 ms | 1.28 µs | ~64,000× |
> | full scan (209) | 5.89 ms | 0.97 ms | ~6× |
> | 1-doc durable write | 58.8 ms | 2.26 ms | ~26× |
> | mixed 1 write + 10 reads | 273 ms | 4.61 ms | ~59× |
> | churn 2000 updates | 441 ms/op | 21.8 ms/op | ~20× |
> 
> The win is hardware-independent (larger on the slow box because the uncached free-list sweep is O(page_count)/commit and CPU-bound). Residual Linux/Mac absolute gaps are fsync-bound (durable writes/churn), not compute; reads track within ~1–2×.
> 
> Against a real 172 MB bit-refinery `.aflite`: point lookup 115 ms → ~1 µs, findings scan 116 ms → 2.5 ms; end-to-end its slowest dashboard pages dropped from ~0.85 s to ~4 ms.
> 
> ## Testing
> 
> New tests included: page-cache put/remove/wholesale-eviction; cached-snapshot ≡ fresh-disk-rebuild across mixed put/delete/dup-key/external-value/empty batches; pinned-snapshot isolation across commits; out-of-band catalog-commit + vacuum invalidation; page-reuse-after-vacuum served correctly.
> 
> Verified green on Linux x64 (0 failed, 0 leaked throughout):
> 
> - `lite-native-test` 105/105 · `lite-cli-test` 49/49 · `lite-core-test` 105/105 · `capi-test` 10/10 · `docstore-test` · `embedded-test` · `lite-go-test` (Go binding)
> - `lib-db-test` 480 passed / 4 skipped
> - full `make zig-unit-test` (2000+ tests) — green except two pre-existing `common/secrets` hermeticity tests that fail only when `OPENAI_API_KEY` is present in the shell env (unrelated to this change; pass under a scrubbed env; CI runs hermetically).
> 
> ## Notes for reviewers
> 
> - The page cache's correctness rests on the cross-process page-immutability argument in the `native.zig` comment (free-page reuse/vacuum gated by the exclusive data-rewrite lock). This is the one claim not exercisable by single-process unit tests — worth a careful read of the lock coverage.
> - The docstore snapshot cache-key assumes vacuum always advances `(commit_sequence, document_root_page)`; verified in the vacuum test path.
> - Space amplification under churn and DB-level open cost (~350–500 ms) are unchanged here — separate follow-ups.
> 
> Draft pending the merge-queue heavy suite (sim/chaos/integration) and e2e.


ajroetker:
```
  Findings

  1. The snapshot cache is unbounded in every production path (main issue). No production caller passes resource_manager into the Lite backend — capi/db.zig:1907, cmd/lite.zig, embedded/db.zig:99, and restore_staging.zig:143 all open without it; only the new unit tests set it. Unlike the page/link caches, which have local 64/16 MiB caps independent of the manager, SharedSnapshot has no local cap: without a manager, createCachedSnapshotLocked caches unconditionally. So a store now pins a full materialized copy of all documents in RAM for the life of the handle (~172 MB for the real bit-refinery file cited in the description), where previously that memory was transient per transaction. It also means the new Prometheus metrics and provisioned budgets for all three lite.* slices will read zero in practice. Suggest either wiring the manager at the embedded/capi/cmd call sites or giving the snapshot cache a local default cap like its siblings.

  2. The key safety comment is attached to the wrong declaration. In native.zig, the multi-line "In-memory cache of encoded pages… Safe because…" doc comment runs directly into the PageLinkInfo doc comment with no blank line, so in Zig it all becomes PageLinkInfo's doc — PageCache itself ends up undocumented. Since this comment is the cross-process immutability argument the PR explicitly asks reviewers to rely on, it should be moved above const PageCache.

  3. FileLocksUnsupported fallback weakens the immutability argument (pre-existing, but worth a note). Both openDataFile and acquireDataRewriteLock silently fall back to unlocked opens on filesystems without lock support. Two unlocked writers were already unsafe, but the page cache changes the failure mode from transiently-inconsistent reads to indefinitely serving stale cached pages after another process reuses them. Consider disabling the page cache (or free-page reuse) when the
  lock acquisition falls back.

  4. Catalog commits invalidate the snapshot cache. Every putCatalogRecord bumps commit_sequence, forcing a full chain re-walk on the next transaction. If the backend's index storage writes catalog records interleaved with document traffic, indexed workloads could lose much of the win. Correctness is fine (there's a test for it); just confirm the target workloads don't interleave heavily.

  5. Nits.
  - readPageAllocForCheckpoint caches the raw page before CRC validation happens in decodePagePayloadAlloc, so a corrupt page gets cached. Behavior stays consistent (disk is equally corrupt, and check() bypasses the cache), but caching only after successful decode would be tidier.
  - The hand-maintained slice lists (snapshotStats inline-for, the metrics list in runtime.zig, the test in provisioned_storage.zig) keep growing in four places; std.enums.values(Slice) would collapse them.
```

zig/pkg/antfly/src/storage/lite/native.zig
```diff
@@ -235,6 +240,266 @@ pub const OwnedCatalogKey = struct {
...
 };
 
+/// In-memory cache of encoded pages, keyed by page id.
+///
+/// Safe because page contents are stable for the lifetime of an open handle:
+/// all in-process page writes flow through `writePage` (which updates the
+/// cache) or `replaceOpenFileWithImage` (which clears it), and other
+/// processes cannot rewrite pages under us — free-page reuse and vacuum
+/// require the exclusive data-rewrite lock, which both our writer-mode
+/// `.lock = .none` read-write handle plus `.aflite.lock` and read-only
```
bpopadiuk: lgtm! codex had a small recommendation about improving comments re: locking logic

Non-blocking clarity nit: this comment might be worth tightening to distinguish the sidecar writer lock from the data-file shared/exclusive locks. My read is that the sidecar lock serializes writers, while read-only shared data-file locks block the exclusive data-rewrite lock used for page reuse/vacuum. Also worth being explicit that the cross-process cache-safety argument depends on file locks being supported, given the existing FileLocksUnsupported fallback.

bpopadiuk (approved):

ajroetker:
```
  Minor notes

  - page_cache_enabled is a plain bool while page_cache_bypass is atomic. It can be flipped by disablePageCacheForUnsupportedLocks (vacuum/free-reuse path) while readers check it, which is technically a data race. The consequence is benign — a racing reader can at worst insert one entry into a cache that's never consulted again, since all subsequent reads check the flag first — but making it an std.atomic.Value(bool) like its neighbor would match the file's own conventions and  satisfy the memory model.
  - The new disable test's last put ("not-admitted-by-nativefile") actually is admitted into the PageCache map, because PageCache.put itself doesn't know about the flag — gating lives at the NativeFile level. The test still passes since it only re-asserts the flag; the comment/name slightly oversell what's checked. Cosmetic.
  - The initNativeSingleFile refactor to pass Store by value with the close_store_owner flag is equivalent to the old pointer-and-undefined dance and its error handling is correct.
```

dovinmu:
```
Bound Lite caches in production paths

- Lite native Handle now creates and owns a default ResourceManager when callers do not provide one.
- That manager is passed into the native docstore and propagated through configureDbOpenOptions.
- Externally supplied managers still win.
- This fixes the main issue: snapshot/page/link cache usage is now budgeted/accounted in normal C API / CLI / embedded Lite handle paths.

Handled unsupported file locks safely

- Lock helpers now return whether the lock was actually supported.
- If Zig reports FileLocksUnsupported, Lite still opens the file, but disables and clears the page/link caches for that NativeFile.
- Cache reads/writes/link lookups now check that flag.
- page_cache_enabled is atomic to avoid a reader race during vacuum/free-page reuse.
- Rationale: page/link cache correctness depends on lock-backed page immutability. If the filesystem cannot provide locks, we preserve compatibility but drop the optimization.

Cleaned up review nits

- Moved the PageCache safety comment onto PageCache.
- Tightened it to explain the sidecar writer lock, read-only shared data-file locks, exclusive data-rewrite lock, and unsupported-lock fallback.
- Removed a misleading test assertion that bypassed NativeFile and wrote directly into PageCache.
- Left catalog-commit snapshot invalidation, CRC-before-cache, and enum-list cleanup for later since they are correctness-safe/non-blocking.
```

PR #336, opened 2026-07-08
Fix HA admin token handling

bpopadiuk:
## Summary

This PR fixes HA operator/runtime-side handling for hot-standby admin actions while preserving the operator's no-Secret-read security boundary.

Key changes:

- Fallback HA admin actions to CLI Jobs when the operator process does not have the configured admin token env var but the Job has token wiring through `admin.envFrom` or `runtime.adminTokenSecretRef`.
- Map HA admin Job tokens from `runtime.adminTokenSecretRef` as a Kubernetes `SecretKeySelector`, without reading Secret values into the operator process.
- Keep operator status probes and typed admin actions on the explicit `spec.highAvailability.admin.tokenEnvVar` path.
- Retry stale/missing failed HA admin Jobs instead of treating old failed direct-admin statuses as permanently terminal.
- Parse JSON HA admin Job results and use completed slot Jobs as evidence for planned-action progress.
- Allow fresh HA retention sentinel behavior.
- Document the runtime Secret ref versus operator env-token boundary in HA/RBAC docs.

## Validation

```text
GOWORK=off GOEXPERIMENT=simd GOCACHE=/private/tmp/antfly-go-cache go test ./controllers/antfly -run 'Test(HAAdminBearerTokenDoesNotReadRuntimeSecretRef|ReconcileHAAdminJobsFallsBackToCLIJobWhenConfiguredAdminTokenEnvVarComesFromRuntimeSecret|ReconcileHAAdminJobsFallsBackToCLIJobWhenConfiguredAdminTokenEnvVarComesFromEnvFrom|ReconcileHAAdminJobsFailsWhenConfiguredAdminTokenEnvVarMissing)'
ok github.com/antflydb/antfly/go/pkg/operator/controllers/antfly 0.701s
```

```text
GOWORK=off GOEXPERIMENT=simd GOCACHE=/private/tmp/antfly-go-cache go test ./controllers/antfly
ok github.com/antflydb/antfly/go/pkg/operator/controllers/antfly 0.601s
```

```text
GOWORK=off GOEXPERIMENT=simd GOCACHE=/private/tmp/antfly-go-cache go test ./api/antfly/v1
ok github.com/antflydb/antfly/go/pkg/operator/api/antfly/v1 0.419s
```

```text
GOWORK=off GOEXPERIMENT=simd GOCACHE=/private/tmp/antfly-go-cache go test ./manifests
ok github.com/antflydb/antfly/go/pkg/operator/manifests 0.272s
```

```text
GOWORK=off GOEXPERIMENT=simd GOCACHE=/private/tmp/antfly-go-cache go test ./...
ok github.com/antflydb/antfly/go/pkg/operator/api/antfly/v1 (cached)
ok github.com/antflydb/antfly/go/pkg/operator/api/inference/v1alpha1 0.247s
?  github.com/antflydb/antfly/go/pkg/operator/bootstrap/antfly [no test files]
?  github.com/antflydb/antfly/go/pkg/operator/cmd/antfly-operator [no test files]
ok github.com/antflydb/antfly/go/pkg/operator/controllers/antfly (cached)
ok github.com/antflydb/antfly/go/pkg/operator/controllers/inference 1.118s
ok github.com/antflydb/antfly/go/pkg/operator/controllers/internal/poddiagnostics 0.269s
ok github.com/antflydb/antfly/go/pkg/operator/internal/webhook/antfly/v1 0.630s
ok github.com/antflydb/antfly/go/pkg/operator/internal/webhook/inference/v1alpha1 1.436s
ok github.com/antflydb/antfly/go/pkg/operator/manifests (cached)
```


ajroetker:
```
• Findings

  - High: fallback jobs are gated on len(admin.EnvFrom) > 0, so runtime-secret-only configs still get stuck. In /private/tmp/antfly-pr-336/go/pkg/operator/controllers/antfly/antflycluster_controller.go:3724, haAdminActionHasEnvFromFallback
    returns false unless admin.EnvFrom is present. But the PR also teaches job creation to inject runtime.adminTokenSecretRef directly into the job env at /private/tmp/antfly-pr-336/go/pkg/operator/controllers/antfly/
    antflycluster_controller.go:7664. For a cluster with admin.tokenEnvVar set, operator env missing, runtime.adminTokenSecretRef set, and no admin.envFrom, direct execution fails locally with errHAAdminTokenEnvMissing, then fallback is skipped,
    and the action is marked failed instead of creating the job that now has enough token wiring to succeed. The fallback predicate should accept the runtime secret path too, or take cluster and check whether haAdminJobTokenEnv(cluster, admin)
    would be non-empty.
 ```

**Comment by bpopadiuk** [comment]
@ajroetker fixed in `69d448904`.

The fallback predicate now receives the cluster and checks whether the fallback Job will have token wiring via either path:

- existing `admin.envFrom`
- `runtime.adminTokenSecretRef` through `haAdminJobTokenEnv(cluster, admin)`

That means the runtime-secret-only configuration you called out no longer gets stuck after local direct execution reports `errHAAdminTokenEnvMissing`; it clears the direct-admin failure state and creates the CLI fallback Job with the token `SecretKeyRef`.

I also added `TestReconcileHAAdminJobsFallsBackToCLIJobWhenConfiguredAdminTokenEnvVarComesFromRuntimeSecret`, which covers the exact case with `admin.tokenEnvVar` set, missing operator env, `runtime.adminTokenSecretRef` set, and no `admin.envFrom`.

Verified with:

```sh
GOWORK=off GOEXPERIMENT=simd GOCACHE=/private/tmp/antfly-go-cache go test ./controllers/antfly -run 'TestReconcileHAAdminJobs(FailsWhenConfiguredAdminTokenEnvVarMissing|FallsBackToCLIJobWhenConfiguredAdminTokenEnvVarComesFromEnvFrom|FallsBackToCLIJobWhenConfiguredAdminTokenEnvVarComesFromRuntimeSecret|RecoversStaleDirectAPIMissingTokenFailureWithEnvFromFallback)'
GOWORK=off GOEXPERIMENT=simd GOCACHE=/private/tmp/antfly-go-cache go test ./controllers/antfly
```

ajroetker:
Codex review note with proposed fix:

I think the latest commit fixes the fallback-job bug, but there is still a secret-boundary/RBAC issue. Status probes now call `haAdminSDKClient(..., true)`, which can fall through to `haAdminRuntimeSecretBearerToken()` and do `r.Get(..., secret)` for `spec.highAvailability.runtime.adminTokenSecretRef`. That means a default operator install with only the runtime token Secret configured will try to read Secret data from the controller.

That conflicts with the existing operator security contract in this repo: the controller comments say Secrets are intentionally not read by the operator, `admin.tokenEnvVar` is documented as avoiding Secret read permissions for the operator, and `manifests/rbac_test.go` explicitly rejects granting `secrets` access. With the current default RBAC, this path should surface forbidden errors during HA admin status observation whenever the operator env token is absent.

Proposed fix: keep runtime Secret handling as a SecretKeySelector reference for runtime pods and fallback Jobs, but do not read `runtime.adminTokenSecretRef` into the operator process. For status probes, require the existing `spec.highAvailability.admin.tokenEnvVar` / operator env injection path. If we intentionally want the operator to read runtime token Secrets, then the PR should update the RBAC, docs, tests, and security comments to make that expanded trust boundary explicit, but I think preserving the no-Secret-read boundary is the safer/default-aligned change.

ajroetker:
Codex follow-up / proposed direction:

I think layered secrets are the cleaner model here, and they line up with what Antfly already partially supports. The runtime already accepts repeated `--secret-store-path` values as primary plus fallback layers, while Colony already generates `cloud-secrets-config` / `secrets.json` for hosted runtime secrets such as `antfly.trusted_principal.secret`.

A better long-term shape would be:

1. Keep the operator out of Secret data. PR 336 should not make the controller read `runtime.adminTokenSecretRef` for status checks; that conflicts with the existing no-Secret-read boundary and RBAC posture.
2. Have Colony generate the HA admin token into a platform-managed Antfly secret layer, for example `antfly.ha_admin.token`, similar to how it writes the trusted-principal secret today.
3. Extend the operator CRD from a single `spec.secretStore` to ordered secret-store layers, then mount each layer and pass each path to Antfly as repeated `--secret-store-path` args.
4. Add HA token resolution by Antfly secret-store key, for example `highAvailability.runtime.adminTokenSecretKey: antfly.ha_admin.token`, instead of requiring the operator to read a Kubernetes Secret.

One caveat: Antfly's layered store treats the first path as primary and later paths as fallbacks. If a user/customer layer comes before a platform layer, user data could override reserved platform keys. For HA/admin/trusted-principal secrets, we probably either want the platform layer to win for reserved keys or explicit reserved-key enforcement for prefixes like `antfly.ha_admin.*` and `antfly.trusted_principal.*`.

Suggested staging:

- Short term in this PR: keep Kubernetes Secret refs only as pod/job env injection and remove operator-side Secret reads.
- Medium term: add ordered secret-store layers to the operator API.
- Colony side: generate the HA admin token into a platform-managed secret layer.
- Runtime side: add explicit HA token lookup by secret-store key.

bpopadiuk:
@ajroetker agreed on preserving the no-Secret-read boundary. I pushed `e1e6e1934` to remove the operator-side runtime Secret read.

What changed:

- Removed the `haAdminRuntimeSecretBearerToken()` path entirely.
- Status probes and typed admin actions now only use the operator process token from `spec.highAvailability.admin.tokenEnvVar`.
- `runtime.adminTokenSecretRef` remains supported only as Kubernetes `SecretKeySelector` wiring for Antfly runtime pods and CLI fallback Jobs.
- Replaced the old runtime-secret-read test with `TestHAAdminBearerTokenDoesNotReadRuntimeSecretRef`.
- Updated the HA and RBAC docs/API comments to state the boundary explicitly.

Validation:

```sh
GOWORK=off GOEXPERIMENT=simd GOCACHE=/private/tmp/antfly-go-cache go test ./controllers/antfly
GOWORK=off GOEXPERIMENT=simd GOCACHE=/private/tmp/antfly-go-cache go test ./api/antfly/v1
GOWORK=off GOEXPERIMENT=simd GOCACHE=/private/tmp/antfly-go-cache go test ./manifests
GOWORK=off GOEXPERIMENT=simd GOCACHE=/private/tmp/antfly-go-cache go test ./...
```

PR #337, opened 2026-07-08
Add prompt-prefix KV caching for native, Metal, and CUDA inference

timkaye11:
## Summary

  Adds prompt-prefix KV caching for inference so repeated/shared prompt prefixes can skip prefill work
  and reuse retained KV blocks.

  - Adds request/API support for `prompt_cache_key`, per-request `prompt_cache`, and
  `cached_prompt_tokens` usage reporting.
  - Adds server config for prompt caching: `enabled`, `max_bytes_mb`, `min_tokens`, and `ttl_ms`.
  - Implements `PromptPrefixCache` with longest page-aligned prefix reuse, TTL expiry, byte-budget
  eviction, and prompt-cache metrics.
  - Extends KV manager/storage runtime with retained-block attach/release support.
  - Reuses cached KV on native CPU, Metal, and CUDA backends.
  - Wires stable `attachIo` lifecycle so Metal device-backed cache hooks do not depend on request-owned
  IO.
  - Disables prompt caching on unsupported/high-risk paths: streaming, multimodal, speculative decode,
  compaction, compiled partition, DeepSeekV4 compressed cache, and Qwen3.5.

  ## Validation

  - `zig fmt --check zig/pkg/inference/src/main.zig zig/pkg/inference/src/server/server.zig zig/pkg/
  inference/src/server/model_manager.zig`
  - `git diff --check`
  - `zig build antfly-client-test`
  - `zig build inference-test -Dmetal=false -- --test-filter "node attachIo"`
  - `zig build inference-test -Dmetal=false -- --test-filter "prompt cache"`
  - `zig build inference-test -Dmetal=false -- --test-filter "run config"`
  - `zig build inference-test -Dmetal=false`
  - Metal smoke with Gemma4 QAT GGUF:
    - first request: no cached tokens
    - repeated request: `cached_prompt_tokens: 32`
    - `/readyz` remained healthy after cache hit


zig/pkg/antfly/src/openapi/generated/antfly_inference_config_openapi/types.zig
```diff
@@ -342,6 +344,10 @@ pub const GenerateRequest = struct {
     stream: ?bool = null,
     /// List of tools (functions) the model can call. Only supported by models with tool_call_format configured.
     tools: ?[]const Tool = null,
+    /// inference-native prompt prefix cache namespace/routing key. Requests with the same key can reuse matching prompt-prefix KV on the same node.
+    prompt_cache_key: ?[]const u8 = null,
```
ajroetker, line 409: Nice!! Should we go ahead and plug this into the resource_manager too?

specs/openapi/inference/api.yaml
```diff
@@ -2476,6 +2476,16 @@ components:
             inference-native KV cache compaction ratio applied after prefill via Attention Matching.
             Selects a subset of keys and fits new values via OLS to preserve attention behavior.
             0.02 = 50x compression, 0.1 = 10x, 0.5 = 2x. Null/omitted = no compaction.
+        prompt_cache_key:
+          type: string
+          maxLength: 256
+          description: |
+            inference-native prompt prefix cache namespace/routing key. Requests with the same key can
```
ajroetker: How does the routing work? Do we need to integrate this with the inference proxy layer at all?

zig/pkg/antfly/src/openapi/generated/antfly_inference_config_openapi/types.zig
timkaye11: ooh yeah thats a good idea! I'll add it

specs/openapi/inference/api.yaml
timkaye11: ah 'routing' might be a stretch, currently the prompt caching is single node w/o any proxy routing atm. i'll update the description

ajroetker:
I'm wondering if we should route to the prompt cache based off of a stable hash that we can route to live cached prefixes using some sort of merkle tree we track on the proxy or something?

timkaye11:
@ajroetker  yeah, I think that'll be a good follow up PR. after chatting with claude , consistent-hash affinity on `prompt_cache_key` seems like a good direction.  Looks like the consistent hashing mechanics already exist in the Go inference proxy , but it's not used for generate/chat models (ie only ready-heavy ops (https://github.com/antflydb/antfly/blob/main/go/pkg/proxy/inference/proxy.go#L804), so we'll need to make some udpates there

ajroetker:
```
• Findings

  - High: .worktrees/pr-337-review/zig/pkg/inference/src/server/server.zig:5472 reads prompt_prefix_cache.stats() from /metrics without taking the same lock used by generation. stats() walks mutable entries and block_hash_entries, while
    generation can append, evict, or swapRemove them. A concurrent metrics scrape during generation can race into invalid memory or inconsistent slices, which is also a DoS risk if metrics are reachable. I’d add a cache-local mutex/snapshot
    method, or make these metrics atomic counters/gauges updated under the cache mutation path.

  - Medium: .worktrees/pr-337-review/zig/pkg/inference/src/runtime/kv/prompt_cache.zig:276 registers errdefer cleanup in an order that frees owned block-id slices before using those slices to release retained KV refs. On allocation failure after
    toOwnedSlice, the later errdefer self.allocator.free(...) runs before releaseRetainedBlocks(...), so the release path reads freed memory. Same issue for storage blocks at .worktrees/pr-337-review/zig/pkg/inference/src/runtime/kv/
    prompt_cache.zig:279. Use a single errdefer { release; free; } block or reverse the registrations so release runs before free.

  - Medium/security: .worktrees/pr-337-review/zig/pkg/inference/src/server/server.zig:2141 enables prompt cache by default per request when the node has it enabled, and .worktrees/pr-337-review/zig/pkg/inference/src/pipelines/generation.zig:2433
    falls back to "" when prompt_cache_key is omitted. That means all no-key callers share a per-model cache namespace. Even though token equality prevents wrong-output reuse, it exposes prompt-presence/timing side channels, amplified by
    cached_prompt_tokens in responses at .worktrees/pr-337-review/zig/pkg/inference/src/server/server.zig:3005. For multi-tenant deployments, require an explicit key or derive one from the authenticated principal; otherwise disable cache when
    the key is absent.
```

bpopadiuk:
@ajroetker @timkaye11 from codex:

```
**Findings**
- Medium: The TTL is documented as an idle TTL, but entries expire on absolute age. Inserts set `expires_at_ms`, while hit paths only update `last_used`, and `expireOld()` evicts only by `expires_at_ms`. A hot prefix will still be dropped every `ttl_ms`, causing periodic full-prefill latency. Refresh expiry on hit/exact duplicate, or rename/docs it as absolute TTL.  
  [[prompt_cache.zig#L317](https://github.com/antflydb/antfly/blob/fee587fdf5c9b30771a559fd342e52d391d44aee/zig/pkg/inference/src/runtime/kv/prompt_cache.zig#L317)](https://github.com/antflydb/antfly/blob/fee587fdf5c9b30771a559fd342e52d391d44aee/zig/pkg/inference/src/runtime/kv/prompt_cache.zig#L317), [[#L249-L251](https://github.com/antflydb/antfly/blob/fee587fdf5c9b30771a559fd342e52d391d44aee/zig/pkg/inference/src/runtime/kv/prompt_cache.zig#L249-L251)](https://github.com/antflydb/antfly/blob/fee587fdf5c9b30771a559fd342e52d391d44aee/zig/pkg/inference/src/runtime/kv/prompt_cache.zig#L249-L251), [[#L540-L557](https://github.com/antflydb/antfly/blob/fee587fdf5c9b30771a559fd342e52d391d44aee/zig/pkg/inference/src/runtime/kv/prompt_cache.zig#L540-L557)](https://github.com/antflydb/antfly/blob/fee587fdf5c9b30771a559fd342e52d391d44aee/zig/pkg/inference/src/runtime/kv/prompt_cache.zig#L540-L557)

- Medium: The production default is `simple`, but that path does linear entry scans plus token-by-token prefix comparison under one cache mutex, and eviction is also linear. That is okay for small caches, but it is not the scalable/most efficient solution for many tenants or many distinct prompt keys. I’d make `block_hash` the production default or clearly mark `simple` as small-cache/debug mode.  
  [[server.zig#L105-L110](https://github.com/antflydb/antfly/blob/fee587fdf5c9b30771a559fd342e52d391d44aee/zig/pkg/inference/src/server/server.zig#L105-L110)](https://github.com/antflydb/antfly/blob/fee587fdf5c9b30771a559fd342e52d391d44aee/zig/pkg/inference/src/server/server.zig#L105-L110), [[prompt_cache.zig#L218-L224](https://github.com/antflydb/antfly/blob/fee587fdf5c9b30771a559fd342e52d391d44aee/zig/pkg/inference/src/runtime/kv/prompt_cache.zig#L218-L224)](https://github.com/antflydb/antfly/blob/fee587fdf5c9b30771a559fd342e52d391d44aee/zig/pkg/inference/src/runtime/kv/prompt_cache.zig#L218-L224), [[#L560-L583](https://github.com/antflydb/antfly/blob/fee587fdf5c9b30771a559fd342e52d391d44aee/zig/pkg/inference/src/runtime/kv/prompt_cache.zig#L560-L583)](https://github.com/antflydb/antfly/blob/fee587fdf5c9b30771a559fd342e52d391d44aee/zig/pkg/inference/src/runtime/kv/prompt_cache.zig#L560-L583)

- Medium: Cache memory is capped per loaded model, but the resource-manager integration only observes usage; it does not enforce or shrink across models. With default `max_bytes_mb = 512` and multiple loaded generator models, enabling this can retain multiple 512 MiB caches even though the resource slice records hard pressure. I’d either derive per-model cache from an aggregate node budget or add a real pressure/shrink callback.  
  [[server.zig#L105-L119](https://github.com/antflydb/antfly/blob/fee587fdf5c9b30771a559fd342e52d391d44aee/zig/pkg/inference/src/server/server.zig#L105-L119)](https://github.com/antflydb/antfly/blob/fee587fdf5c9b30771a559fd342e52d391d44aee/zig/pkg/inference/src/server/server.zig#L105-L119), [[model_manager.zig#L727](https://github.com/antflydb/antfly/blob/fee587fdf5c9b30771a559fd342e52d391d44aee/zig/pkg/inference/src/server/model_manager.zig#L727)](https://github.com/antflydb/antfly/blob/fee587fdf5c9b30771a559fd342e52d391d44aee/zig/pkg/inference/src/server/model_manager.zig#L727), [[resource_manager.zig#L283-L298](https://github.com/antflydb/antfly/blob/fee587fdf5c9b30771a559fd342e52d391d44aee/zig/pkg/antfly/src/storage/resource_manager.zig#L283-L298)](https://github.com/antflydb/antfly/blob/fee587fdf5c9b30771a559fd342e52d391d44aee/zig/pkg/antfly/src/storage/resource_manager.zig#L283-L298)
```

ajroetker:
[codex]

Review findings from the latest pass:

- Medium: The prompt-cache “node-wide” memory cap is not actually enforced across already-active model caches. `runtimeConfig()` splits `max_bytes_mb` by active cache count, but only the current model is reconfigured on request setup, so cache A can keep its old 512 MiB budget while cache B/C get smaller shares. Aggregate retained KV can exceed the documented node-wide cap until each older model is requested again. The resource-manager hook only observes pressure; it does not reject or force shrink.
  - `.worktrees/pr-337-review/zig/pkg/inference/src/server/server.zig:112`
  - `.worktrees/pr-337-review/zig/pkg/inference/src/server/server.zig:2647`
  - `.worktrees/pr-337-review/zig/pkg/inference/src/server/model_manager.zig:1059`
  - `.worktrees/pr-337-review/zig/pkg/antfly/src/storage/resource_manager.zig:283`

- Medium/security, if this API is multi-tenant or exposed to untrusted clients: cache isolation is based entirely on the client-supplied `prompt_cache_key`. Any caller that guesses or reuses another caller’s key can share the namespace and observe cache-hit signals via latency and `usage.cached_prompt_tokens`. The keyless default is fixed, which is good, but production multi-tenant isolation should derive/prefix the namespace from the authenticated principal server-side, not trust the request body as the full isolation boundary.
  - `.worktrees/pr-337-review/zig/pkg/inference/src/server/server.zig:2180`
  - `.worktrees/pr-337-review/zig/pkg/inference/src/server/server.zig:2206`
  - `.worktrees/pr-337-review/zig/pkg/inference/src/server/server.zig:3727`

Verification: `zig build inference-test --summary failures` passed locally: 2012 selected, 2001 passed, 11 skipped.

ajroetker:
[codex]

Additional review finding from the deeper cache/storage pass:

- Medium: GPU prompt-cache memory is under-accounted even for a single active cache. The Metal/CUDA cache path builds the pool config with `store_cpu_bytes` left at its default `true`, then `ensureStorage()` creates both the cache manager’s host pool and a separate `KvStorageRuntime` with that same config. Each retained block can therefore allocate CPU KV bytes in both places, and the backend hook may also retain device-side KV slots, but `estimateBytes()` only charges one KV copy. So `max_bytes_mb` can substantially understate actual retained memory on Metal/CUDA, independent of the cross-model budget issue already posted. I’d either set the metadata-only manager/storage side to `store_cpu_bytes = false` when the device hook is authoritative, or make the estimator count the manager host bytes, storage host bytes, and device-resident bytes explicitly.
  - `.worktrees/pr-337-review/zig/pkg/inference/src/server/server.zig:2628`
  - `.worktrees/pr-337-review/zig/pkg/inference/src/runtime/kv/prompt_cache.zig:171`
  - `.worktrees/pr-337-review/zig/pkg/inference/src/runtime/kv/prompt_cache.zig:181`
  - `.worktrees/pr-337-review/zig/pkg/inference/src/runtime/kv/prompt_cache.zig:549`
  - `.worktrees/pr-337-review/zig/pkg/inference/src/runtime/kv/pool.zig:302`

PR #339, opened 2026-07-09
Fix HA failover recovery from live Kind validation

bpopadiuk:
## Summary

Fixes the HA runtime/operator issues uncovered while completing the live e2e Kind validation for hot standby mode, plus the follow-up CI regressions caught after rebasing onto the latest operator/runtime work.

The local rollout exercised real `AntflyCluster` CRs, pods, Services, fencing Lease, promotion, stale-primary writes, former-primary rejoin/reseed, and backup/restore through the promoted public route. These checks exposed several gaps that were not covered by the earlier fixture/unit/envtest evidence.

## Issues uncovered and fixed

- Promoted standbys did not fully rewire into primary mode after promotion, leaving stale routing/write-gate behavior and incomplete primary-side HA state.
- Former primaries could receive rejoin requests without durably recording the promotion fence receipt, so stale-primary write checks could fail to reject direct writes after the old pod returned.
- Backup through the live public route could fail when the provisioned backup path tried to open a second writer for a DB that already had a live cached writer.
- The operator could preserve stale or failed direct-admin evidence and re-plan former-primary reseed incorrectly.
- Former-primary reseed could retry `SeedStandby` with a new manifest after the promoted primary already had an active healthy seed slot, causing `BaseBackupSlotInUse`.
- Rejoin result matching was too strict for demote/assessment-only paths and for promotion fork-LSN evidence.
- Operator CI exposed stale seed-LSN expectations after the runtime changed fresh base-backup manifests to target `primaryLSN + 1`.
- Operator CI also exposed former-primary fencing detection that ignored durable promotion receipts and a rejoin result matcher that rejected the admin server's authoritative `former_last_lsn`.
- Zig e2e exposed promoted-standby write/owner-job gates validating `expected_identity` against the old standby identity before honoring a durable promoted-primary handoff.
- HA storage tests exposed table-scoped mirror fixtures that were not opening DBs with the matching identity namespace, plus rejoin admin tests that used synthetic fence receipts conflicting with the local fence store.

## Validation

Live Kind rollout evidence remains the production-readiness evidence for the Colony checklist; the additional local test runs below validate the CI failures addressed in this PR.

- Deployed runtime image from this branch into `kind-colony-cloud-test`.
- Verified Kubernetes Lease acquisition and promotion receipt.
- Verified public route switched to the promoted `standby-a` pod.
- Verified direct stale writes to the old primary Service and pod IP return `409 fenced primary rejects writes`.
- Verified former-primary rejoin/reseed status ends with `ReseedFormerPrimary` succeeded and no recurring `BaseBackupSlotInUse`.
- Verified table backup, cluster backup, manifest presence, overwrite restore, and removal of a post-backup-only key through the promoted public route.
- `GOCACHE=/private/tmp/antfly-go-cache GOMODCACHE=/private/tmp/antfly-go-mod-cache make test` in `go/pkg/operator`: passed.
- `ZIG_LOCAL_CACHE_DIR=/private/tmp/antfly-zig-local ZIG_GLOBAL_CACHE_DIR=/private/tmp/antfly-zig-global zig build ha-test --summary all`: passed, 275 tests.
- Native e2e binary build: `zig build -Dedition=full install`: passed.
- `ANTFLY_BIN=./zig-out/bin/antfly uv run --project e2e/antfly pytest -q -x e2e/antfly/test_standby.py::test_standby_streams_public_writes_restarts_and_rejects_writes`: passed.

### DIFF
--- a/go/pkg/operator/controllers/antfly/antflycluster_controller.go
+++ b/go/pkg/operator/controllers/antfly/antflycluster_controller.go
@@ -3444,6 +3444,7 @@ func (r *AntflyClusterReconciler) updateStatus(ctx context.Context, cluster *ant
 		if err := r.observeHAFencingStatus(ctx, cluster); err != nil {
 			return err
 		}
+		r.observeHAFormerPrimaryFenceStatus(ctx, cluster)
 		r.updateHAStatusAndConditions(cluster)
 		if haAdminStatusErr != nil {
 			setHACondition(
@@ -3539,6 +3540,7 @@ func (r *AntflyClusterReconciler) updateStatus(ctx context.Context, cluster *ant
 	if err := r.observeHAFencingStatus(ctx, cluster); err != nil {
 		return err
 	}
+	r.observeHAFormerPrimaryFenceStatus(ctx, cluster)
 	r.updateHAStatusAndConditions(cluster)
 	if haAdminStatusErr != nil {
 		setHACondition(
@@ -5147,8 +5149,15 @@ func (r *AntflyClusterReconciler) applyHADirectRejoinJobResult(cluster *antflyv1
 	}
 	action.AdminResult = haRejoinAdminActionResult(result)
 	if !haActionHasRequiredAdminResult(*action) {
-		action.AdminResult = nil
-		return false
+		if haActionKind(action.Kind) == haActionDemoteFormerPrimary &&
+			strings.TrimSpace(action.AdminResult.RejoinAction) != "" &&
+			strings.TrimSpace(action.AdminResult.FormerNodeID) != "" {
+			// DemoteFormerPrimary records the assessment step. A valid assessment
+			// may reject rejoin before any rewind/reseed execution result exists.
+		} else {
+			action.AdminResult = nil
+			return false
+		}
 	}
 	if cluster.Status.HAStatus.FormerPrimary == nil {
 		cluster.Status.HAStatus.FormerPrimary = &antflyv1.HAFormerPrimaryStatus{NodeID: action.StandbyName}
@@ -5239,14 +5248,16 @@ func haDirectRejoinResultMatchesAction(result haRejoinJobResult, status *antflyv
 	if action.StandbyName != "" && result.FormerNodeID != action.StandbyName {
 		return false
 	}
-	if action.TargetLSN > 0 && result.ForkLSN != action.TargetLSN {
-		return false
-	}
-	if action.ObservedLSN > 0 && result.FormerLastLSN != action.ObservedLSN {
-		return false
+	if haActionKind(action.Kind) == haActionDemoteFormerPrimary {
+		if action.ObservedLSN > 0 && result.FormerLastLSN != action.ObservedLSN {
+			return false
+		}
 	}
 	if action.RetainedFromLSN > 0 && result.RetainedFromLSN != action.RetainedFromLSN {
 		return false
+	}
+	if haActionKind(action.Kind) == haActionDemoteFormerPrimary {
+		return true
 	}
 	if status != nil && status.LastPromotion != nil {
 		promotion := status.LastPromotion
@@ -7371,11 +7382,10 @@ func haRejoinResultMatchesRequiredAdminResult(action antflyv1.HAPlannedActionSta
 	if action.StandbyName != "" && result.FormerNodeID != action.StandbyName {
 		return false
 	}
-	if action.TargetLSN > 0 && result.ForkLSN != action.TargetLSN {
-		return false
-	}
-	if action.ObservedLSN > 0 && result.FormerLastLSN != action.ObservedLSN {
-		return false
+	if haActionKind(action.Kind) == haActionDemoteFormerPrimary {
+		if action.ObservedLSN > 0 && result.FormerLastLSN != action.ObservedLSN {
+			return false
+		}
 	}
 	if action.RetainedFromLSN > 0 && result.RetainedFromLSN != action.RetainedFromLSN {
 		return false
@@ -7384,6 +7394,9 @@ func haRejoinResultMatchesRequiredAdminResult(action antflyv1.HAPlannedActionSta
 }
 
 func haFormerPrimaryActionSucceededWithPromotionEvidence(status *antflyv1.HAStatus, action antflyv1.HAPlannedActionStatus) bool {
+	if haActionKind(action.Kind) == haActionDemoteFormerPrimary {
+		return haAdminActionSucceededWithEvidence(action)
+	}
 	return haAdminActionSucceededWithEvidence(action) &&
 		haRejoinResultMatchesPromotion(status, action, action.AdminResult)
 }
@@ -7402,10 +7415,7 @@ func haRejoinResultMatchesPromotion(status *antflyv1.HAStatus, action antflyv1.H
 	if result.TargetTimelineID != promotion.NewTimelineID || result.TargetEpoch != promotion.NewEpoch {
 		return false
 	}
-	if promotion.SwitchLSN != 0 && result.ForkLSN != promotion.SwitchLSN {
-		return false
-	}
-	if promotion.RequiredLSN != 0 && result.ForkLSN != promotion.RequiredLSN {
+	if expectedForkLSN := haPromotionObservedLSN(promotion); expectedForkLSN != 0 && result.ForkLSN != expectedForkLSN {
 		return false
 	}
 	if haActionKind(action.Kind) == haActionRewindFormerPrimary &&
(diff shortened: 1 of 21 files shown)


ajroetker:
```
• Findings

  1. Security/DoS: rejoin accepts and persists caller-supplied fence receipts before binding checks
     .worktrees/pr-339-review/zig/pkg/antfly/src/storage/ha/http_admin.zig:820 now calls fence_store.recordReceipt(fence) immediately after parsing the request body. recordReceipt validates only receipt structure/monotonicity, so a caller to /
     admin/v1/ha/rejoin/* can durably install a fabricated fence before the receipt is checked against node_id, local identity, or assessment outcome. If HA admin auth is unset, this is unauthenticated; if set, any HA admin token holder can
     fence the primary through an assessment-style endpoint. Persist only after validating the receipt matches the requested former primary and expected local identity, or avoid mutating fence state from rejoin assessment.

  2. Promoted standby does not update the live HTTP server’s internal HA executor
     .worktrees/pr-339-review/zig/pkg/antfly/src/data/runtime.zig:2542 creates a new internal HA server and writes its executor to self.api_server_cfg, but ApiHttpServer already holds a copied config and dispatches from
     self.cfg.ha_internal_executor in .worktrees/pr-339-review/zig/pkg/antfly/src/api/http_server.zig:2159. After promotion, the live public server can still have a null/stale internal executor, so /internal/ha/... routes may not serve the
     promoted primary until restart.

  3. Promotion closes a standby handle still owned by the swarm caller
     .worktrees/pr-339-review/zig/pkg/antfly/src/data/runtime.zig:2510 closes the standby object in DataServer, but in the swarm path that object is stack-owned as ha_standby and still has a deferred close at .worktrees/pr-339-review/zig/pkg/
     antfly/src/swarm/runtime.zig:1003. A successful promotion can leave the outer owner closing an already-closed/undefined standby later. Transfer ownership explicitly or have DataServer only detach the pointer while the owner handles close.
```

ajroetker:
[codex]

Suggested design to address the review findings:

1. Fence receipt persistence

Make rejoin receipt handling non-mutating by default. `/rejoin/assess`, `/rejoin/rewind`, and `/rejoin/reseed` should parse the supplied receipt as evidence only. Before any durable write, validate that the receipt is bound to the request and local node: `node_id == receipt.old_primary_id`, request identity matches the receipt parent identity, the promoted identity matches the expected target timeline, and the local role/context is allowed to record the receipt.

Best security posture: do not call `recordReceipt` from rejoin at all; use the receipt only for assessment. If durable receipt recording is required for stale-primary rejection, persist only after validation and only when the receipt fences this exact local node. I would also avoid exposing a generic `recordReceipt(receipt)` API for external evidence. Prefer something like `recordVerifiedReceipt(receipt, expected_binding)` so every caller must supply the expected local binding.

2. Live internal executor rewire

Avoid copying mutable HA executors into `ApiHttpServerConfig` as the long-lived source of truth. Give `ApiHttpServer` a shared mutable executor holder, for example an `HAExecutors` cell with `admin` and `internal` executors, owned by `DataServer`. Promotion then updates that cell once, and `dispatchHaRoutes` always reads the current executor.

Lower-impact fix: after promotion, update both `self.api_server_cfg.ha_internal_executor` and `self.http_server.?.cfg.ha_internal_executor`. That fixes the stale live route, but keeps two sources of truth. The shared holder is less fragile.

3. Standby ownership / double close

Make HA handles either owned by `DataServer` or borrowed by `DataServer`, never both. The current swarm path stack-owns `ha_standby` and defers close, while promotion closes the same object from `DataServer`.

Preferred design: transfer ownership of HA handles into `DataServer` when it may perform role transitions, and have the swarm caller clear its local optional so it does not defer-close transferred handles. Alternative: keep them borrowed and have promotion only detach the pointer while the original owner closes later. Since promotion opens `ha_promoted_primary` and changes runtime role, I would choose owned handles.

Implementation order I would suggest:

1. Fix HA handle ownership first.
2. Replace executor copying with a shared mutable executor holder.
3. Make rejoin receipt persistence explicit and binding-checked, or remove it from rejoin entirely.
4. Add tests for: forged receipts cannot be persisted through rejoin; promoted primary serves internal HA routes through the live API server without restart; promotion does not double-close a standby opened by swarm/runtime setup.

Security rule to preserve: only the promotion/fence acquisition path should create durable fencing authority. Rejoin endpoints can consume fence evidence, but should not silently install it unless the local-node binding has been proven.

bpopadiuk:
@ajroetker thanks for the careful review. I pushed `f7ef948a4` with mitigations for the three findings.

1. Rejoin receipt persistence: removed the eager `fence_store.recordReceipt` from rejoin handling. `/rejoin/assess` is now non-mutating even if the caller supplies a receipt. For typed mutating actions (`/rejoin/rewind` and `/rejoin/reseed`), the receipt is first validated against the request binding (`node_id == receipt.old_primary_id` plus parent identity/timeline/epoch checks). Durable persistence now goes through `recordVerifiedReceipt(receipt, expected_binding)`, and only records when the receipt fences the exact local primary node. I also made the old generic store write private so external evidence has to pass an explicit binding check.

2. Live internal executor rewire: added a live update path on `ApiHttpServer` and call it after promotion rebuilds the promoted-primary internal HA server. I chose the lower-impact fix here rather than the shared executor holder refactor, because it directly fixes the stale live-route bug without reshaping the public server config ownership in this PR. There is now a regression test that starts the API server before promotion, observes `/internal/ha/replication/identify` as unavailable, promotes the standby, and verifies the same live server serves the internal HA route without restart.

3. Standby ownership / double close: added explicit standby ownership handoff for the swarm-owned optional. `DataServer` now consumes the owned standby through the owner slot during promotion and clears that optional, so the swarm defer sees `null` and does not close the promoted standby a second time. I kept the handle stack-owned by swarm setup, but made the transfer/consume point explicit at promotion time.

Added coverage for forged receipts not being persisted through rejoin, assessment staying read-only, live internal HA route availability after promotion, and standby ownership consumption. Local validation passed:

- `zig build ha-test --summary all` (`277 passed; 0 failed; 0 leaked`)
- `zig build -Dedition=full install --summary all`
- `GOWORK=off GOEXPERIMENT=simd GOCACHE=/private/tmp/antfly-go-cache go test ./...` under `go/pkg/operator`

bpopadiuk:
@ajroetker running the tests locally:

[local-test-run.txt](https://github.com/user-attachments/files/29899130/local-test-run.txt)

here's just the one that's failing in CI:

```zsh
➜  ~ cd /Users/boris/dev/github/antflydb/antfly
gh pr checkout 339
git rev-parse HEAD
Already on 'boris/pr346-live-kind-fixes'
Your branch is up to date with 'origin/boris/pr346-live-kind-fixes'.
Already up to date.
f7ef948a413a19270b276d81c82fd14bd1b3fab6
➜  antfly git:(boris/pr346-live-kind-fixes) cd /Users/boris/dev/github/antflydb/antfly/zig
ZIG_LOCAL_CACHE_DIR=/private/tmp/antfly-zig-local-pr339 \
ZIG_GLOBAL_CACHE_DIR=/private/tmp/antfly-zig-global-pr339 \
zig build -fincremental -Dedition=full install --summary all
Build Summary: 12/12 steps succeeded
install success
├─ install antfly success
│  └─ compile exe antfly Debug native success 1m MaxRSS:9G
│     ├─ options cached
│     ├─ options (reused)
│     ├─ WriteFile tokenizer.json cached
│     ├─ options cached
│     ├─ options cached
│     └─ run exe patch_sentencepiece_proto (sentencepiece_proto) success 190ms
│        ├─ compile exe patch_sentencepiece_proto ReleaseSafe native success 3s MaxRSS:346M
│        ├─ run exe protoc-zig (sentencepiece_proto_raw) success 418ms
│        │  └─ compile exe protoc-zig Debug native success 1s MaxRSS:310M
│        └─ run exe protoc-zig (sentencepiece_proto_raw) (+1 more reused dependencies)
└─ install ../go/pkg/antfly/src/metadata/antfarm/ cached

➜  zig git:(boris/pr346-live-kind-fixes) cd /Users/boris/dev/github/antflydb/antfly
SKIP_BUILD=1 RUN_INFERENCE_E2E=0 \
ZIG_LOCAL_CACHE_DIR=/private/tmp/antfly-zig-local-pr339 \
ZIG_GLOBAL_CACHE_DIR=/private/tmp/antfly-zig-global-pr339 \
scripts/ci/zig-e2e-base-linux.sh \
  e2e/antfly/test_resolution.py::test_multinode_autograph_resolves_promotes_and_hydrates_entities
Darwin Boriss-MacBook-Pro.local 25.5.0 Darwin Kernel Version 25.5.0: Mon Apr 27 20:41:12 PDT 2026; root:xnu-12377.121.6~2/RELEASE_ARM64_T6050 arm64
uv 0.11.28 (Homebrew 2026-07-07 aarch64-apple-darwin)
antfly dev (zig runtime)
================================================================================================ test session starts =================================================================================================
platform darwin -- Python 3.14.6, pytest-9.0.2, pluggy-1.6.0
rootdir: /Users/boris/dev/github/antflydb/antfly/zig/e2e/antfly
configfile: pyproject.toml
collected 1 item

e2e/antfly/test_resolution.py .                                                                                                                                                                                [100%]

================================================================================================= 1 passed in 8.21s ==================================================================================================
➜  antfly git:(boris/pr346-live-kind-fixes)
```

PR #342, opened 2026-07-10
Exclude-set live-doc filtering and cached WAL retention snapshots

dovinmu:
> ## What
> 
> Two independent fixes for pathologies found while debugging the VectorDBBench 1M results, one on the query path and one on the ingest path.
> 
> ### 1. Express the broad live-doc constraint as a small exclude set
> 
> Any table that has ever deleted a document permanently loses the all-docs-visible fast path. Every dense query then materializes the **entire live-doc set** as a positive filter (all doc-nums mapped to vector ids) and pushes it through the HBC search — O(table size) per query, forever. Measured on a 768d/1M table: ~16.5ms of constraint derivation plus ~15ms of in-search filter overhead **per query**, from a single tombstone. (The June live-doc work cached the set derivation, but the per-query clone/projection/mapping remained.)
> 
> Fix: derive the **complement** instead — the ordinals a query must not return (every deleted doc, plus docs created after the read generation) — memoized per identity generation with negative caching of the overflow case (>65536 entries falls back to the include path), and express the broad constraint as that exclude set. One tombstone becomes a one-entry exclusion instead of a million-entry include filter; a tombstone whose doc had no vector costs nothing at all.
> 
> Scope note: the exclusion representation is deliberately restricted to executors that use ordinals directly (dense, match-all). Sparse/text include filters are projected to index doc-nums, which *also* implicitly drops stale rows of re-indexed docs — an exclusion of non-visible ordinals cannot express that, so those paths keep the include representation (caught by `db chunked generated dense and sparse embeddings search as parent results`).
> 
> Measured (50k OpenAI case, fresh tombstone): constraint derivation 16.5ms/1M-row filter → **0.5ms/empty exclusion**, identical recall. On the 1M benchmark table this was the difference between a flat ~40–150ms p99 band and the real recall/latency curve (0.98 recall @ ~10ms p99).
> 
> ### 2. Cache WAL retention snapshots for maintenance and pressure checks
> 
> `wal.snapshotRetention` re-reads the checkpoint index and the current segment, and fstats every retained segment. Write-pressure enforcement, maintenance scoring, and stats snapshots all invoked it fresh on **every maintenance step under the backend lock**. Under sustained ingest, the moment async index catch-up lags slightly, retained WAL crosses its hard limit and the enforcement re-scans after every flush in its inner loop — monopolizing the lock and starving both the foreground writers and the async appliers whose progress would release the retention. Loads stalled in waves (insert timeouts around 70–500k docs) and dense catch-up wedged mid-replay; every stall `sample` across multiple runs shows the same frames on top.
> 
> Fix: serve those callers from a 250ms cache on the backend (invalidated inside the pressure flush loop so it observes its own progress), and debounce the enforcement pass itself to the same interval. Retention limits are approximate by nature, so bounded staleness is immaterial. A follow-up could replace the TTL cache with incremental retention accounting updated on rotate/checkpoint events.
> 
> Measured (768d/1M ingest, write sync, concurrency 1, unpaced, quiet machine): before, wedged at ~67k docs; after, **full 1M load in 1,845s (~31 min) with zero stalls and the dense index fully caught up at load end** — vs 3,408s (~57 min) for the previously-only-reliable config (client-side paced) plus its post-load catch-up tail.
> 
> ## Why draft
> 
> Rebasing onto current main surfaced 3 `lib-db-test` failures (510 pass / 4 skip) that look like interactions with the recently-landed visibility/hit-total machinery (`requires_full_candidate_visibility_filter`, total relations) rather than with the original bases these fixes were validated on:
> 
> - [ ] `db non chunked search paths apply broad live doc filter` — expects hit-total relation `.exact`, gets `.gte` under the exclusion path
> - [ ] `db batch persists per-index applied sequence watermark` — possibly a retention-cache interaction with watermark persistence
> - [ ] one further `TestExpectedEqual` failure to identify
> 
> Both fixes passed the full suite (498/0) on their original bases; the reconciliation above is the remaining work before marking ready.
> 
> ## Context
> 
> Full investigation trail with per-stage batch profiles, stall samples, and A/B methodology: `#devlogs > fable-antfly-yolo-v2` and `#devlogs > why-load-so-slow` (2026-07-09/10).
>


ajroetker:
[codex]

Review findings from `./.worktrees/pr-342-review`:

1. **High: WAL hard-limit enforcement is now time-throttled.**
   `enforceWalRetentionHardPressure()` calls `walRetentionEnforceDue()` before it even snapshots/checks hard pressure (`zig/pkg/antfly/src/storage/lsm_backend.zig:4113`). Because `walRetentionEnforceDue()` suppresses checks for 250ms (`:3972`), a fast second commit can cross `wal_hard_limit_segments` and skip the forced foreground flush/checkpoint. This weakens a configured hard disk-retention bound and the existing `lsm backend hard wal pressure forces foreground checkpoint on commit` test now fails at `:6684`.

   Suggested fix: keep the short interval for best-effort/background maintenance, but do not gate the hard-limit path before checking current pressure. For foreground write pressure, either always take a fresh/cached snapshot and enforce when `walRetentionOverHardLimit(retention)` is true, or split the throttle so it only applies to soft-limit/background scoring. The hard limit should remain a hard bound.

2. **Medium: WAL retention cache is stale after checkpoint/reset/retire operations.**
   `snapshotMaintenanceStats()` now reads cached retention (`lsm_backend.zig:1580`), but `resetWalAfterManifestCheckpoint()` resets WAL at `:3118` and `maybeCheckpointWalAfterManifestPublish()` retires covered segments at `:2126` without invalidating the new cache. As a result, maintenance stats and resource-accounting can report retained WAL after it has been checkpointed away. Several existing checkpoint/retention tests fail with retained segments/bytes still nonzero after checkpoint/reset.

   Suggested fix: call `invalidateWalRetentionCacheLocked()` after every WAL-retention-changing operation: `wal_mod.reset`, `wal_mod.retireCoveredSegments`, and any durable-boundary/checkpoint path that can make retained segments disappear. Then recompute resource-manager usage from a fresh snapshot or have `syncTrackedWalRetentionUsageCurrentLocked()` also refresh/update the cached snapshots consistently.

3. **Medium: dense-search total-hit relation regresses from `exact` to `gte` for small tombstone exclude sets.**
   The new broad live-doc fast path represents visibility as `.include = .all` plus a small `.exclude` set (`search_exec.zig:2467`). For dense search that leaves `native_constraints.positive_filter == false`, so candidate counting falls back to full index active count (`search_exec.zig:12007`) and then marks the result `gte` when `k` is smaller than that bound (`:12210`). Existing DB integration now fails at `db.zig:33503`, expecting the page to remain exact after filtering one tombstone.

   Suggested fix: teach dense/sparse candidate-bound logic about negative-only live filters. For this broad-live exclude case, bound the candidate count by active_count minus the mapped exclude count where safe, or carry an explicit “visibility-filtered exclude-only” flag and compute total-hit relation after postprocess when the returned candidate window proves the visible set is exhausted. The old exactness guarantee should not be lost just because the live filter is expressed as a complement.

Validation run:

```sh
zig build db-test -- --test-filter "non-visible doc set" --test-filter "broad live doc filter" --test-filter "wal pressure" --test-filter "retained wal" --test-filter checkpoint
```

Result: 25 passed, 10 failed. The new non-visible set unit tests pass; the failures are integration regressions around WAL retention freshness/enforcement and dense visibility exactness.

dovinmu:
other issues also found
```
  1. High: Dense exclusion fast path bypasses TTL visibility.
     The complement only encodes identity visibility, but the shortcut disables live_filter_doc_set, which also
     performs TTL filtering. On a TTL-enabled table with a tombstone or historical-generation constraint, dense
     top-k may select expired vectors, remove them only during postprocessing, and return an underfilled page
     despite valid lower-ranked documents existing.
     Relevant code: search_exec.zig:2459-2481, doc_identity.zig:760-795, db.zig:14071-14090.

  2. Medium: The cache is also stale when WAL retention grows, not only after checkpoint/reset.
     The posted review covers stale-high snapshots after retention shrinks. The opposite direction is also
     missing: appendWalForMutable() writes/rotates the WAL without invalidating or updating
     cached_wal_retention. An immediate maintenance snapshot can therefore still report segment 0 or an older
     segment. This explains:
      - db batch persists per-index applied sequence watermark
      - db dense auto bulk finish wakes weak-sync replay and publishes visibility after catch-up

     Any fix should update/invalidate the cache after appends and rotations as well as after reset/retire
     operations.

  3. Low/performance: Fresh retention scans still occur after every WAL append.
     syncTrackedWalRetentionUsageCurrentLocked() still calls both snapshotRetention and snapshotReplayRetention
     after every append, even when no resource manager exists, because the resource-manager check happens after
     the scans. Those fresh results also do not refresh the new cache—so the implementation pays for fresh scans
     while other callers can still receive stale data.
```

dovinmu:
> **Codex:** Pushed `ea51cd275` with the review follow-ups: TTL-safe broad visibility filtering, exact dense totals for exhaustive negative-only visibility windows, coordinated WAL retention state across append/retire/reset/replay, and unthrottled foreground hard-limit checks.
>
> Added TTL and WAL cache-coherence regressions. Validation is green: the focused review suite passed 35/35, the three previously failing DB tests now pass, and `zig build lib-db-test` passed 514 tests with 4 expected skips and 0 failures.

Issue #346, opened 2026-07-12
Zig engine: cache query-text embeddings across concurrent table queries (multi-table fan-out re-embeds the same string N times)


markhayden:
## Setup

Bakin fans one user query out to ~11 tables in parallel (`POST /db/v1/tables/{t}/query`, hybrid `full_text_search` + `semantic_search`, v0.2.0-rc.18, darwin-arm64, bge-small on Metal).

## Observed

- Per-table engine `took`: 0–8ms.
- Wall time per hybrid table query: ~250ms (single table, warm).
- 11-table parallel fan-out: ~1.0s wall — the query-text embed jobs appear to serialize on the inference queue, and the *identical string* is embedded once per table.

FTS-only fan-out over the same tables: ~1ms/table, wall ~50ms — the delta is entirely query-time embedding.

## Ask

A tiny LRU keyed by `(model, normalized_text)` for **query-time** embeddings (not documents) would collapse N identical embeds per fan-out into one, taking multi-table search from ~1s to ~300ms without any API change. Alternatively (or additionally): accept a pre-computed query vector on the public query API so clients can embed once and fan out.

Happy to test a build.

ajroetker:
fwiw we do have the ability to take a pre-computed query vector but an lru cache here is a good idea and got dropped from the go version, I'll work on that!

markhayden:
The bounded-singleflight query-embedding cache commits (2026-07-12, in main and rc.21) look like they address exactly this. We can't cleanly measure cache hits from the client side — if the singleflight covers the concurrent multi-table fan-out case this issue describes, feel free to close.

PR #348, opened 2026-07-12
Cache query embeddings with bounded singleflight

ajroetker:
## Summary

- add a process-local query embedding LRU with idle TTL, byte-bounded admission, and concurrent miss coalescing
- key cache entries by effective embedding configuration, credential identity/generation, exact input, and a server-derived security domain
- add an injectable hard cache budget, API/data-server metrics, focused tests, and the broader design in `zig/INFERENCE_CACHING.md`

This addresses #346 by placing singleflight before provider pacing and embedded inference queue acquisition. Equivalent embedding configurations can share results across tables, while authenticated principals, anonymous traffic, and internal callers remain isolated.

Templated and multimodal queries intentionally bypass this cache because templates may resolve mutable remote content. Inference-node L2 and prompt-KV participation in the shared budget remain follow-up integrations; prompt caching is not present on `main` yet.

## Validation

- `zig build --summary failures`
- `zig build root-test --summary failures -- --test-filter "query embedding cache"`
- `zig build root-test --summary failures -- --test-filter "cache budget atomically"`
- `zig build lib-data-runtime-test --summary failures -- --test-filter "data runtime health metrics include replay debt"`
- `zig fmt --check` on all changed Zig files
- `git diff --check`


ajroetker:
https://github.com/antflydb/antfly/issues/346

PR #349, opened 2026-07-12
Productionize Florence-2 Metal KV-cache decoding

timkaye11:
## Summary

Productionizes Florence-2 Metal inference with correct SDPA layouts, default-on KV-cache decoding, safer model loading, and a representative performance gate.

## Changes

- Fixes token-major QKV/SDPA correctness and malformed-buffer crashes.
- Adds preallocated KV-cache decoding and single-pass Metal SDPA.
- Corrects token limits, UTF-8 output, CLI JSON, and API defaults.
- Validates complete Florence bundles before benchmarking.
- Aligns HA E2E coverage with current fenced-primary semantics.

## Validation

- ReleaseSafe Metal build and focused host/Metal tests pass.
- Strict-resident OCR and captioning pass on the complete Q4_K bundle.
- Cached/full output parity is exact.

| Tokens | Cached p50 | Full p50 | Speedup |
| ---: | ---: | ---: | ---: |
| 32 | 8.04s | 10.35s | 1.29× |
| 64 | 12.06s | 18.48s | 1.53× |


ajroetker:
Can you remove claude from the commits?

specs/openapi/inference/api.yaml
```diff
@@ -2123,6 +2125,8 @@ components:
         max_tokens:
           type: integer
```
ajroetker, line 2126: can you make sure `make generate` get's run for this

ajroetker:
```

  1. [P1] Omitted max_tokens breaks readers with smaller model limits — server.zig:4917 (https://github.com/timkaye11/antfly/blob/45fc125a8c8f07f6bc9aaf2d9d81bd2be3aa0cf5/zig/pkg/inference/src/server/server.zig#L4917)

     validateReadMaxTokens(null) converts an omitted value into an explicit 256. The vision reader then adds its decoder prefix and rejects the request when that exceeds the model’s configured max_length. For example, the supported Xenova/trocr-
     base-printed configuration has max_length: 20, so a normal request without max_tokens now fails.

     This affects HTTP reads, direct reads, and image extraction. Keep omission distinct for execution—use 256 only for queue estimation—or clamp an omitted default to the model limit while continuing to reject an explicitly oversized value.

  2. [P2] The new Metal verification script always reports missing GGUF files on macOS — verify_florence2_metal_kv_cache.sh:111
     (https://github.com/timkaye11/antfly/blob/45fc125a8c8f07f6bc9aaf2d9d81bd2be3aa0cf5/zig/pkg/inference/scripts/verify_florence2_metal_kv_cache.sh#L111)

     macOS/BSD find supports neither -maxdepth nor -quit. Because stderr is discarded, the command substitution is empty and every valid model directory is rejected as missing weights. Since Metal verification primarily targets macOS, use a Bash
     glob such as "$model_dir"/*.gguf or another BSD-compatible check.
```

PR #359, opened 2026-07-14
Productionize bounded multi-model inference

### LINKED ISSUE #351: Metal: prepare_embedding_table dominates every warm embed op (~28:1 vs lookup) — bge-small query embed ~0.77s warm on Apple Silicon
# Query-text embedding ~0.77s warm for bge-small on Metal (~40x expected; dominated by prepare_embedding_table)

**Version:** v0.2.0-rc.18 (macOS arm64 / Apple Silicon, swarm mode, `--preload-model embedder:BAAI/bge-small-en-v1.5`)

## Summary

A single-word `semantic_search` query against a small table costs ~0.77s **warm** (1.2s cold) end-to-end, where bge-small inference for a one-sentence input should be ~10–20ms on this hardware. Profiling the embedder thread during batch work shows nearly all time under `prepare_embedding_table` rather than the actual lookup/forward pass — it appears the embedding weight table is re-prepared per operation instead of being prepared once per loaded model.

## Measurements

Direct HTTP against the engine, single node, otherwise idle:

| Request | Time |
|---|---|
| `{"full_text_search":{"match_all":{}},"limit":3}` (48-doc table) | 67ms |
| `{"semantic_search":"budget","indexes":["embeddings"],"limit":3}` — cold | 1.24s |
| same — warm (immediately repeated) | 0.77s |

Because each per-table query embeds the query text independently, a cross-table search fanning out to N tables queues N embeds on the single embedder thread — with 11 tables that is ~8s of serialized embedding for ONE user query, which blows any reasonable per-request deadline. Under any concurrent write/enrichment load, hybrid queries time out engine-wide.

## Profile (macOS `sample`, embedder thread at 100% during batch embed)

```
974 inference.managed_embedder.ManagedEmbedder.embedDenseParts
  974 inference.managed_embedder.embedWithEntryParts
    974 … embedBatchWithEntry
      974 swarm.runtime.localAntflyEmbedDenseTexts
        974 pipelines.embedding.EmbeddingPipeline.embed
          89 architectures.bert.embeddings
          : 89 ops.metal_compute.MetalCompute.embeddingLookupOp
          :   85 termite_metal_decode_runtime_prepare_embedding_table   ← ~95% of the op
          :   3  termite_metal_decode_runtime_embedding_lookup_prepared_device
```

`prepare_embedding_table` dominating `embedding_lookup` ~28:1 inside `embeddingLookupOp` suggests the prepared table is not cached across ops/chunks.

## Prior art / scope

Distinct from (but compounding with) the open inference issues: #346 tracks the multi-table fan-out re-embedding the same query string N times, and #310 tracks cold-start compile persistence + the dead `Node.embed_cache` for repeat queries. This issue is about the WARM cost of a single embed op: `prepare_embedding_table` dominating `embedding_lookup` ~28:1 inside one `embeddingLookupOp`, i.e. the prepared weights table appears to be rebuilt per op rather than cached per loaded model. Fixing this shrinks every embed (queries, enrichment, backfills); #346/#310 then multiply the win.

## Ask

Cache the prepared embedding table per loaded model (prepare once at load/preload, reuse per op).

Happy to provide full profiles or run patched builds against our workload.

timkaye11:
closes #351 

 ## Summary

  - Add bounded model lifecycle management with keep-alive eviction, load caps, and singleflight cold
  loads.
  - Harden Metal and shared ONNX execution with bounded provider pools, cooperative locking, and safe
  embedding caching.
  - Add stable `503` capacity responses, operational metrics, generated API updates, and fail-on-skip
  Metal CI.

  ## Validation

  - Metal hardware: 4/4 passed, 0 skipped
  - Lifecycle/backend: 12/12 passed
  - Configuration: 30/30 passed
  - Release build, OpenAPI, Go, and TypeScript checks passed

  ## Follow-up

  Replace the conservative request watermark with per-request model leases to avoid rare unnecessary
  `503`s during overlapping multi-model requests.


ajroetker:
[codex] Static code review found two merge regressions around the new bounded-capacity contract.

1. **Restore 503 handling in the high-level Go inference client**

`go/pkg/sdk/inference.go` currently ignores the generated `JSON503` response in `Embed`, `EmbedMultimodal`, `EmbedJSON`, `Chunk`, `ChunkMedia`, `SparseEmbed`, and `SparseEmbedJSON`. A normal capacity rejection is consequently returned as an `unexpected JSON response` or generic unexpected-status error, while generate/rerank/rewrite still report `service unavailable` consistently.

Proposed fix:

- Restore the `if resp.JSON503 != nil { ... }` branches in all seven helpers, ideally through a shared helper so endpoint handling cannot drift again.
- Preserve the stable `MODEL_CAPACITY_REACHED` / `SERVICE_UNAVAILABLE` code and, if practical, expose `Retry-After` through a typed retryable error rather than only an error string.
- Restore `TestClient_ServiceUnavailable` and cover both capacity codes plus the `Retry-After` header.

2. **Regenerate the Termite Go API from the updated inference schema**

`specs/openapi/inference/config.yaml` now sets bounded defaults for `max_loaded_models` and `max_concurrent_requests`, marks both with `x-omitempty: false` / `x-omitzero: false`, and documents immediate backpressure. `go/pkg/termite/api.gen.go` still contains the old unlimited/queued descriptions and `omitempty,omitzero` tags, and its embedded Swagger is therefore stale.

Proposed fix:

- Regenerate `go/pkg/termite/api.gen.go` using its declared `//go:generate` source (`specs/openapi/inference/config.yaml`). The resulting fields should preserve explicit zero values and the embedded spec should describe the new behavior.
- Add a small serialization regression asserting that explicit zero values for both limits remain present in JSON.
- If defaults 10/32 are also intended for the Go Termite runtime, apply them during configuration construction because OpenAPI defaults are not automatically applied by Go JSON unmarshalling. Otherwise, clarify in the schema that those defaults are Zig-runtime-only.

I reviewed this statically and did not run the test suite.

timkaye11:
```
=================================== FAILURES ===================================
__________ test_autoscaling_finalizes_shard_split_from_size_threshold __________
```

^ e2e-base failure

PR #360, opened 2026-07-14
Add quality-aware server-side PDF OCR ingestion

dovinmu:
> ## Summary
> 
> Extends the server-side document-ingestion work in #335 with quality-aware, per-page PDF OCR. This PR targets `aj/align-full-e2e` so it lands on top of #335 rather than creating a competing PDF pipeline.
> 
> - Accept raw PDFs from inline artifact payloads or fetched URLs.
> - Extract embedded text page-by-page, then selectively OCR deficient pages or OCR every page when explicitly configured.
> - Render only the requested page at configurable DPI with a pixel safety cap.
> - Use a layout/table-preserving OCR prompt and retain whichever result—embedded or OCR—scores better.
> - Automatically chunk, embed, and populate named/default full-text indexes for semantic, BM25, and hybrid retrieval.
> - Preserve page/chunk offsets and expose per-page OCR method, quality, failure, selection, and provenance through manifests and APIs.
> - Cover inline and URL PDFs, scanned and born-digital documents, garbled text, scanned tables, inline JPEG captioning, page-N rendering, chunking, and retrieval.
> 
> Docsaf remains a separate client of Antfly. It only submits Antfly's ingest configuration; Antfly does not depend on docsaf.
> 
> ## Configuration and operations
> 
> OCR fallback thresholds, mode (`auto` or `always`), render DPI, and pixel limits are configurable in the server-side `document_extraction` hierarchy. Transient failures remain retryable, permanent page failures are isolated, and usable embedded text is retained when OCR fails. Manifests summarize attempted, selected, retained, and failed pages and identify failed page numbers.
> 
> The PR also preserves artifact-backed full-text/vector enrichments during table creation, keeps table-status responses compatible with generated SDKs, and makes docsaf's default chunk enrichment populate the public/default lexical surface.
> 
> ## Corpus hardening
> 
> OHR-Bench exposed native PDF paths that could crash or silently produce zero units before OCR. This PR now also:
> 
> - fixes recursive xref trailer ownership and tolerates out-of-range unused xref entries without accepting malformed syntax;
> - walks the real page tree when `/Count` is missing, stale, or zero, and reports a genuine empty tree explicitly;
> - fixes Type1/Type3 encoding lifetime and allocation-failure cleanup;
> - correctly decodes ToUnicode `bfrange` UTF-16BE strings, including ligatures and surrogate handling; and
> - fixes OCR quality flag arithmetic overflow.
>


ajroetker:
[codex] I found two issues in the PDF changes, with proposed fixes:

1. **The `lib-pdf-test` target does not execute the PDF module's tests.**

   `zig/lib/pdf/pdf_test_root.zig` only calls `std.testing.refAllDecls(antfly_pdf)`. Tests declared in the imported `antfly_pdf` module are not registered by that wrapper. On this PR head, `zig build lib-pdf-test --summary all` reports only `1/1 tests passed`, which is the wrapper test; none of the tests in `mod.zig`, `reader.zig`, `syntax.zig`, or `render.zig` run.

   I verified this by temporarily making `pdf_mod` the `b.addTest` root. That exposed 15 compilation errors, including stale five-argument calls to `extractTextRunsFromContentAppend` after its signature changed, plus the fixture package-path errors. This means the new page-number, DPI, pixel-guard, and scanned-table tests currently provide no runtime coverage.

   **Proposed fix:** make the real PDF module the test root (or create test roots that directly own/import the PDF source files in a way that registers their tests), then move/expose the fixtures within the module package path and fix the existing test compilation errors. As a regression guard, `zig build lib-pdf-test --summary all` should report the actual PDF test count rather than one wrapper test.

2. **`scalePatternRuns` scales tiling patterns in both page space and pattern space.**

   The code scales `pattern_matrix`, but also scales `pattern_bbox`, `pattern_{x,y}_step`, and every tile-content run. The renderer first maps world coordinates through the inverse `pattern_matrix`, so scaling both sides changes the tile's frequency instead of only increasing raster resolution.

   I reproduced this with a 4-unit tile whose left half is red and right half is white. At 2x DPI, the expected output is `red x4, white x4`, repeated twice across 16 pixels. The current scaling produces `red x8, white x8` instead. This affects any OCR render above 72 DPI, including the default 150 DPI, when a page uses tiling patterns.

   **Proposed fix:** scale the target path/stroke/clip geometry because that is in page space, and left-multiply the pattern-to-page transform by the DPI scale (equivalent to scaling the six `pattern_matrix` components here). Keep `pattern_bbox`, `pattern_x_step`, `pattern_y_step`, and tile contents unchanged because they remain in pattern space. For shading-pattern mode, continue scaling the shading geometry that is consumed directly in page space. A renderer test with an asymmetric/repeating tile at 72 and 144 DPI would catch this coordinate-space regression.

ajroetker:
[codex] Production-readiness follow-up

The happy path is working, but I don't think this is production-ready yet. I found the following blockers and proposed fixes:

1. **[P1] `ocr.mode = "always"` discards embedded fallback text**
   - In both PDF extraction paths, forced OCR replaces the embedded text and regions with empty values before OCR runs:
     https://github.com/antflydb/antfly/blob/69277956f36649252d46ab9eadbdcfa71ef03939/zig/pkg/antfly/src/storage/db/enrichment/document_extraction.zig#L1122-L1128
   - If OCR fails or scores worse, there is no embedded text left to retain.
   - **Proposed fix:** always extract embedded text/regions; use `force_ocr` only to enqueue OCR, then compare OCR against the preserved embedded result.

2. **[P1] Large scanned PDFs have quadratic behavior**
   - The generated-unit cache performs linear scans on insert and lookup:
     https://github.com/antflydb/antfly/blob/69277956f36649252d46ab9eadbdcfa71ef03939/zig/pkg/antfly/src/storage/db/enrichment/enrichment_runtime.zig#L3255-L3281
   - Each OCR page also constructs a new PDF reader and traverses the document again:
     https://github.com/antflydb/antfly/blob/69277956f36649252d46ab9eadbdcfa71ef03939/zig/lib/pdf/src/mod.zig#L67-L75
   - **Proposed fix:** replace the cache scan with a `StringHashMap` keyed by unit ID, and retain one parsed reader/flattened page index for extraction and page rendering.

3. **[P1] The rendered-pixel limit does not bound transient memory**
   - Page streams/images are decoded before the pixel check. A permitted 40 MP page is already about 160 MB of raw RGBA, followed by PNG, base64, and JSON copies.
   - A single request larger than `max_bytes` is still accepted because the batching condition only flushes a non-empty batch:
     https://github.com/antflydb/antfly/blob/69277956f36649252d46ab9eadbdcfa71ef03939/zig/pkg/antfly/src/storage/db/enrichment/enrichment_runtime.zig#L2816-L2827
   - **Proposed fix:** validate page geometry before decoding render runs, cap all intermediate output, reserve/account render + PNG + base64 + request memory, and reject an individual request exceeding the batch limit.

4. **[P1] Retry behavior differs between execution paths**
   - The synchronous path converts every producer error, including transient 429/network/5xx failures, into `failed_ocr`:
     https://github.com/antflydb/antfly/blob/69277956f36649252d46ab9eadbdcfa71ef03939/zig/pkg/antfly/src/storage/db/db.zig#L18212-L18221
   - The async batch path correctly propagates retryable/control errors.
   - **Proposed fix:** share the same retry classification helper and propagate transient errors in every path; isolate only permanent page failures.

5. **[P1] The PDF test target is false-green**
   - `zig build lib-pdf-test` runs only the wrapper's `refAllDecls` test:
     https://github.com/antflydb/antfly/blob/69277956f36649252d46ab9eadbdcfa71ef03939/zig/lib/pdf/pdf_test_root.zig#L17
   - It reports 1/1 passing, while making the real PDF module the test root exposed 15 compile errors.
   - **Proposed fix:** import/run the actual test declarations and make this a required PR check.

6. **[P2] Positive-but-stale PDF page counts silently omit pages**
   - Any positive `/Count` is trusted without validating the page tree:
     https://github.com/antflydb/antfly/blob/69277956f36649252d46ab9eadbdcfa71ef03939/zig/lib/pdf/src/reader.zig#L1209-L1229
   - Reproduction: changing a two-page fixture from `/Count 2` to `/Count 1` makes `pageCount()` return 1.
   - **Proposed fix:** flatten/validate the page tree once and derive both page count and lookup from actual `Page` leaves.

7. **[P2] Pattern rendering is incorrect above 72 DPI**
   - The pattern matrix and pattern-local geometry are both scaled, changing tile frequency.
   - **Proposed fix:** scale device/page placement only; keep tile-local coordinates and steps in their original pattern space. This is also described in the earlier [codex] comment.

UX/API follow-ups:
- Validate OCR provider configuration when the table/enrichment is registered instead of failing during ingestion.
- Expose trigger reasons as an array and quality records as structured objects rather than JSON encoded inside strings.
- State explicitly that the CLI OCR flags only apply with `--create-table`.
- Add metrics for render/provider latency, request bytes, retry class, batch/queue depth, and memory pressure.

Validation performed:
- ReleaseFast build passed.
- Go `docsaf` tests passed.
- Focused PDF OCR E2E passed.
- `git diff --check` passed.
- `lib-pdf-test` currently reports only the single wrapper test.

My recommendation is to address the P1 items before merge, then add a large scanned-PDF stress test and transient-provider retry test as release gates.

dovinmu:
[codex testing on Office QA benchmark]

There are two distinct bugs, both reproducible using the same URL-only ingestion flow with:

```
 ANTFLY=http://127.0.0.1:8080

  curl -fsS -X POST "$ANTFLY/db/v1/tables/pdf_native_repro" \
    -H 'Content-Type: application/json' \
    -d '{
      "num_shards": 1,
      "indexes": {
        "document_units": {
          "type": "graph",
          "source": {
            "kind": "artifact",
            "artifact": "document_units_v1",
            "path": "$.edges[*]",
            "format": "extraction_relation"
          },
          "artifact": {
            "name": "document_units_v1",
            "kind": "asset",
            "field": "url",
            "content_type": "application/json",
            "producer_json": {
              "type": "document_extraction",
              "config": {
                "route_preset": "mixed_files",
                "source": {
                  "filename_field": "filename",
                  "content_type_field": "mime_type",
                  "checksum_field": "sha256"
                },
                "ocr": {
                  "enabled": true,
                  "render_dpi": 150,
                  "quality": {
                    "min_content_chars": 50
                  },
                  "config": {
                    "provider": "antfly",
                    "model": "antflydb/Florence-2-base"
                  }
                }
              }
            }
          },
          "edge_types": [{"name": "mentions"}]
        }
      }
    }'
```
  Once the table is ready, insert the two URL-only source rows:
```
  curl -fsS -X POST "$ANTFLY/db/v1/tables/pdf_native_repro/batch" \
    -H 'Content-Type: application/json' \
    -d '{
      "sync_level": "full_index",
      "inserts": {
        "1941_01_treasurybulletin.pdf": {
          "id": "1941_01_treasurybulletin.pdf",
          "url": "https://fraser.stlouisfed.org/files/docs/publications/tbulletin/1941_01_treasurybulletin.pdf",
          "filename": "1941_01_treasurybulletin.pdf",
          "mime_type": "application/pdf",
          "sha256": "54324f8cc1ed2af1b51dfcc65e43f4811eaad1f1790772f73a8485b986fddb0a",
          "doc_type": "source_document"
        },
        "2024_09_treasurybulletin.pdf": {
          "id": "2024_09_treasurybulletin.pdf",
          "url": "https://fraser.stlouisfed.org/files/docs/publications/tbulletin/2024_09_treasurybulletin.pdf",
          "filename": "2024_09_treasurybulletin.pdf",
          "mime_type": "application/pdf",
          "sha256": "844e4d30fa916e15dbde168344815c3eaebeefecf0e62809441dfaf32793db7c",
          "doc_type": "source_document"
        }
      }
    }'
```
  Inspect the resulting artifact states directly:
```
  curl -fsS \
    "$ANTFLY/db/v1/tables/pdf_native_repro/documents/1941_01_treasurybulletin.pdf/artifacts/document_units_v1?detail=raw" |
    jq .

  curl -fsS \
    "$ANTFLY/db/v1/tables/pdf_native_repro/documents/2024_09_treasurybulletin.pdf/artifacts/document_units_v1?detail=raw" |
    jq .
```
  Expected failures:

  - 1941: 78 units, 24 OCR attempts, 21 UnsupportedPdfRendering failures, and three successful Florence calls.
  - 2024: extraction aborts with MissingTable and produces zero units.

ajroetker:
I didn't delete the branch, it auto-closed when I merged the base PR

PR #366, opened 2026-07-15
Add quality-aware server-side PDF OCR ingestion

dovinmu:
## Summary

Adds end-to-end server-side PDF ingestion to Antfly. One artifact configuration accepts inline PDF bytes or a URL, fetches and parses the PDF inside Antfly, emits page units, selectively OCRs, chunks, embeds, and populates lexical plus semantic indexes—without external preprocessing or intermediate JSON.

## What changed

- **Quality-aware OCR:** configurable fallback heuristics and forced OCR, embedded-vs-OCR quality selection, table/layout-aware prompts, Florence's `<OCR>` default, and prompt-echo/trivial-output rejection. Inline JPEG captioning is also supported.
- **Bounded rendering and transport:** 150-DPI target rendering adapts to configurable dimension/pixel caps (never below a safe 72-DPI floor), emits compressed PNGs, and applies inference-specific limits to Antfly-generated data images while retaining remote-content limits for external URLs.
- **Failure handling and provenance:** records per-page extraction/OCR method, selection, requested/effective DPI, dimensions, encoded size, timing, concrete failure stage, and retryability. Failed OCR keeps usable embedded text and remains observable/reprocessable; malformed OCR options are rejected and producer credentials remain write-only in status responses.
- **Indexing correctness and performance:** atomically provisions artifact-backed indexes, feeds chunks into full-text search, batches synchronous OCR pages through the shared Reader/Florence path, and batches document embeddings across page boundaries with bounded item/byte caps and persisted telemetry. Exact source counts propagate through runtime status so strict coverage stays correct across sync barriers, restore, and restart.
- **PDF compatibility and safety:** improves multi-page rendering/caching and common fonts, CMaps, forms, images, predictors, and encrypted PDFs while bounding decode filters, traversal, decompression, canvas, pattern, and transparency work. Unit and public `/merge` E2Es cover inline/URL, born-digital, scanned, garbled, table, paged OCR, hybrid retrieval, and restart coverage paths.

## Notes

Replaces #360 after #335 merged and includes current `main`, including its Florence optimizations. DocsAF is a client/configurator that uses Antfly; Antfly does not depend on DocsAF.


ajroetker:
```
unit-test
+- run test 181 pass, 1 fail (182 total)
error: 'src.reader.test.apply explicit mask alpha uses mask alpha or grayscale channel' failed:
       expected 11, found 255
       /home/runner/_work/_tool/zig/0.16.0/x64/lib/std/testing.zig:118:17: 0x16d1514 in expectEqualInner__anon_73259 (std.zig)
                       return error.TestExpectedEqual;
                       ^
       /home/runner/_work/_tool/zig/0.16.0/x64/lib/std/testing.zig:83:5: 0x17a098e in expectEqual (pdf_test_root.zig)
           return expectEqualInner(T, expected, actual);
           ^
       /home/runner/_work/antfly/antfly/zig/lib/pdf/src/reader.zig:11163:5: 0x17a09d5 in test.apply explicit mask alpha uses mask alpha or grayscale channel (pdf_test_root.zig)
           try std.testing.expectEqual(@as(u8, 11), rgba[3]);
           ^
failed command: /mnt/cache/antfly/zig/zig-base/29382115987-1/zig-local/o/fd31d3bbc7004a0bc67f3b278967f961/test --cache-dir=/mnt/cache/antfly/zig/zig-base/29382115987-1/zig-local --seed=0x8a1a06e4 --listen=-```

File: zig/pkg/antfly/src/api/tables.zig [hunk 1]
```diff
@@ -1835,42 +1837,15 @@ fn buildCanonicalIndexConfigValue(
...
-}
-
 fn projectInlineEnrichmentConfigsInTableStatusJson(alloc: std.mem.Allocator, encoded: []const u8) ![]u8 {
-    var parsed = try std.json.parseFromSlice(std.json.Value, alloc, encoded, .{});
-    defer parsed.deinit();
-    var owned = try cloneJsonValueAlloc(alloc, parsed.value);
-    defer deinitJsonValue(alloc, &owned);
-    try projectInlineEnrichmentConfigsInTableStatusValue(alloc, &owned);
-    return try std.json.Stringify.valueAlloc(alloc, owned, .{ .emit_null_optional_fields = false });
+    return try alloc.dupe(u8, encoded);
```
bpopadiuk: Codex (GPT-5): [P1] Do not return unredacted producer configuration from table status. This change removes the projection and preserves each inline enrichment object, including `producer_json`. The new docsaf OCR flag accepts arbitrary reader-provider JSON, and reader configs support fields such as `api_key` and `credentials_path`, so a literal credential stored in the OCR config is now echoed by the list/get table-status APIs to every caller allowed to inspect table metadata. Please keep provider config write-only, redact secret-bearing fields recursively, or expose a typed non-secret summary rather than the serialized producer payload.

zig/lib/pdf/src/reader.zig
```diff
@@ -4204,11 +4752,19 @@ fn applyStreamFilterAlloc(
...
+        var flate_buffer: [std.compress.flate.max_window_len]u8 = undefined;
+        var decompress: std.compress.flate.Decompress = .init(&in, .zlib, &flate_buffer);
+        var inflated_list = std.ArrayList(u8).empty;
+        defer inflated_list.deinit(alloc);
+        var read_buf: [64 * 1024]u8 = undefined;
+        const max_decoded_stream_bytes: usize = 256 * 1024 * 1024;
+        while (true) {
+            const n = try decompress.reader.readSliceShort(&read_buf);
+            if (n == 0) break;
+            if (inflated_list.items.len > max_decoded_stream_bytes -| n) return error.DecodedStreamTooLarge;
```
bpopadiuk, line 5151: Codex (GPT-5): [P1] Apply the decoded-stream limit to every filter stage, not just Flate. The same dispatcher sends LZW and RunLength streams to unbounded `ArrayList` decoders below, and filter arrays can chain those decoders after this check. With the default 100 MiB remote-download allowance, RunLength alone can expand input by roughly 64x, and chained filters can amplify it again; these PDF-library allocations also occur before the document-extraction working-set tracker can reject them. Enforce one cumulative decoded-byte budget on the output of every filter/predictor stage.

zig/pkg/antfly/src/storage/db/enrichment/document_extraction.zig
```diff
@@ -311,13 +638,94 @@ pub fn parseConfig(alloc: Allocator, raw: []const u8) !Config {
...
+    if (ocr.get("mode")) |mode| {
+        if (mode != .string) return error.InvalidDocumentExtractionConfig;
+        config.ocr_mode = if (std.mem.eql(u8, mode.string, "auto"))
+            .auto
+        else if (std.mem.eql(u8, mode.string, "always"))
+            .always
+        else
+            return error.InvalidDocumentExtractionConfig;
+    }
+    if (intField(ocr, "render_dpi")) |dpi| {
```
bpopadiuk: Codex (GPT-5): [P2] Reject a present option whose JSON type is wrong instead of treating it as absent. `intField`/`floatField` return null for non-numeric values, so configs such as `"render_dpi":"600"` or string-valued quality thresholds pass table validation and silently use defaults. The same pattern makes `"enabled":"false"` fall through `boolField(... ) orelse true`, unexpectedly enabling OCR and its model calls. Because this configuration is already JSON embedded inside `producer_json`, accepting typos with different semantics is especially hard for SDK users to diagnose. Please distinguish missing from malformed fields (and ideally reject unknown OCR keys) during create/update validation.

bpopadiuk (changes requested):
Codex (GPT-5): I reviewed this draft specifically for API-consumer UX and security. I’m requesting changes for two security boundaries and one configuration-contract issue. The local PDF suite passes, and the focused metadata test confirms the new status-response behavior discussed inline.

specs/openapi/antfly/metadata.yaml
```diff
@@ -2043,6 +2043,27 @@ components:
         chunk_count:
           type: integer
           description: Number of indexable chunks derived from the units.
+        ocr_attempted_count:
```
ajroetker, line 2075: I don't think this should be "ocr" specific (for instance it should also apply to auto image embedding of pages OR ocr) same for all these fields, a lot of them apply to chunks too

zig/pkg/inference/src/pipelines/reading.zig
```diff
@@ -1442,12 +1469,37 @@ fn tokenTensorToU32(cb: *const ComputeBackend, allocator: std.mem.Allocator, tok
...
+        std.log.info("read-profile correlation_id={s} phase={s} elapsed_ms={d:.3}", .{ correlation_id, phase, nsToMs(nowNs() - start_ns) });
+    } else {
+        std.log.info("read-profile phase={s} elapsed_ms={d:.3}", .{ phase, nsToMs(nowNs() - start_ns) });
+    }
 }
 
 fn logReadProfileStep(phase: []const u8, step: usize, seq_len: usize, elapsed_ns: u64) void {
     if (!readProfileEnabled()) return;
-    std.log.info("read-profile phase={s} step={d} seq_len={d} elapsed_ms={d:.3}", .{ phase, step, seq_len, nsToMs(elapsed_ns) });
+    if (active_read_profile_correlation_id) |correlation_id| {
```
ajroetker: @timkaye11 can you review the inference changes

zig/pkg/antfly/src/main.zig
```diff
@@ -22,6 +22,7 @@ const antfly_client = @import("antfly-client");
 const platform = @import("antfly_platform");
 
 const antfly_cloud_binary = "antfly-cloud";
+const recommended_server_fd_limit: std.posix.rlim_t = 4096;
```
ajroetker: I don't think this file should reference std.posix? Seems like maybe this should be in lib/platform and the resource_manager abstraction

zig/pkg/inference/src/server/server.zig
```diff
@@ -1213,6 +1213,7 @@ pub const Node = struct {
         const results = try reader.readBatch(image_datas, .{
             .prompt = request.prompt,
             .max_tokens = max_tokens,
+            .correlation_id = request.correlation_id,
```
ajroetker: This is typically called a "trace_id" and we should probably name it that

zig/pkg/antfly/src/chunking/chunk.zig
```diff
@@ -323,6 +353,13 @@ test "append artifact fields stores unit-local and document-global provenance" {
         .page_rotation = 90,
         .extraction_method = "pdf_text",
         .confidence = 0.94,
+        .ocr_attempted = true,
```
ajroetker: See my other comment but I don't think these fields should be ocr specific but should be more generic, maybe have a little convo with fable and codex about making this more adaptable to satisfy embedding multipage directly, ocr, document extraction, captioning, chunking for text/audio/images

zig/pkg/antfly/src/main.zig
ajroetker: [codex] Please move the OS mechanism behind antfly_platform.process, for example ensureFileDescriptorSoftLimitAtLeast(target: u64), and keep only the 4096 policy in startup code. The existing storage ResourceManager does not seem like the right owner unless it is deliberately expanded beyond memory and storage budgets. Also, isServerSubcommand already diverges from the dispatch block: it omits lite and ha, so those processes never receive the check. Ideally server command classification should have one source of truth.

zig/pkg/inference/src/pipelines/reading.zig
ajroetker: [codex] @timkaye11 Could you review the inference portion, particularly the new Florence batch path and serial fallback, batch byte/image limits, the lifetime and nesting of the thread-local profiling context, and whether local-only diagnostic propagation is intentional?

zig/pkg/antfly/src/chunking/chunk.zig
ajroetker: [codex] I would normalize this as provenance.source plus provenance.derivations[], instead of continuing to add sibling ocr_* fields. Each derivation can have a common operation, status, selected flag, parent references, output descriptor, confidence, and failure. Render dimensions and DPI belong to a render derivation, while OCR quality belongs to an OCR derivation. Chunks and embeddings should point to their parent unit/artifact through the same mechanism. The current payload emits the OCR values in three places, which makes future captioning, transcription, and multimodal embedding awkward and prone to drift. A pragmatic boundary for this PR would be to keep the internal OCR-specific Zig fields temporarily but map them into the generic public schema before merging.

zig/pkg/inference/src/server/server.zig
ajroetker: [codex] Looking further upstream, this value is a stable prefix of the source SHA-256, so trace_id would overstate its semantics. Please either name it profile_source_id/source_fingerprint throughout and log it explicitly as such, or introduce a real per-execution trace ID and propagate that through the extraction and inference calls. correlation_id is too vague, but a stable content fingerprint should not be presented as a distributed trace identifier.

specs/openapi/antfly/metadata.yaml
ajroetker: [codex] Concrete proposal: replace the top-level ocr_* fields with processing_stages: []DocumentArtifactStageSummary. Each stage should identify an operation (extract, render, OCR, caption, transcribe, chunk, or embed) and expose common counts such as input, attempted, completed, selected, retained_source, failed, and output. Failures should identify a generic unit_id plus an optional locator, rather than only PDF page numbers. OCR-specific quality metrics can remain nested in the operation details. This gives chunking and direct page/image embedding the same observability contract.

ajroetker:
[codex] e2e-base is still red: https://github.com/antflydb/antfly/actions/runs/29602361560/job/87957590465. The artifact restart test reaches produced/skipped/terminal-failed = 1/1/1, but strict coverage remains incomplete because the group is reported stale/config-mismatched. Please root-cause that state transition rather than widening the wait. The autoscaling split failure may be independent, but it also needs a clean rerun or diagnosis before merge.

ajroetker:
[codex] Re-review of the current head (01c28ba62ff5a0d9ad6fa5bc96ee3b26b5242e43): there have been no code commits since the previous review, and the existing review threads remain unresolved.

1. [P1] The decoded-stream limit is still bypassable during ASCII85 allocation.

The fix checks ASCII85 output only after ascii85DecodeAlloc returns. Inside the decoder, every z expands to four bytes with no budget check: https://github.com/antflydb/antfly/blob/01c28ba62ff5a0d9ad6fa5bc96ee3b26b5242e43/zig/lib/pdf/src/reader.zig#L4878-L4880 and https://github.com/antflydb/antfly/blob/01c28ba62ff5a0d9ad6fa5bc96ee3b26b5242e43/zig/lib/pdf/src/reader.zig#L4925-L4955. A filter chain can RunLength-decode roughly 4 MiB into 256 MiB of z, then make ASCII85 attempt a 1 GiB allocation before DecodedStreamTooLarge is returned. Pass the cap into ascii85DecodeAlloc and check it before every append.

2. [P1] Predictor parameters use unchecked arithmetic from an untrusted PDF.

Columns, Colors, and BitsPerComponent are multiplied directly before casting to usize: https://github.com/antflydb/antfly/blob/01c28ba62ff5a0d9ad6fa5bc96ee3b26b5242e43/zig/lib/pdf/src/reader.zig#L5114-L5120. Large integers can overflow or trap instead of returning a parser error. Use checked multiplication/addition, reject unsupported component counts and bit depths, and ensure row_len is bounded by decoded.len and the decoded-stream budget before allocation.

3. [P2] Credential redaction can silently delete legitimate table schema fields.

redactInlineEnrichmentProducerConfigs recursively removes every object key named producer_json from the entire serialized TableStatus, rather than only enrichment configurations: https://github.com/antflydb/antfly/blob/01c28ba62ff5a0d9ad6fa5bc96ee3b26b5242e43/zig/pkg/antfly/src/api/tables.zig#L1840-L1871. Document schemas permit arbitrary JSON Schema properties, so a legitimate field named producer_json disappears from status responses: https://github.com/antflydb/antfly/blob/01c28ba62ff5a0d9ad6fa5bc96ee3b26b5242e43/specs/openapi/antfly/schema.yaml#L35-L62. Scope redaction to indexes.*.enrichments[*] and table-level artifact enrichments, with a regression test for a document schema property named producer_json.

Fix assessment:

- Malformed OCR option types: fixed correctly, although unknown keys remain accepted.
- Producer credential leak: the leak is fixed, but the redaction is overly broad as described above.
- Decoded stream bounds: incomplete; RunLength and LZW are bounded during growth, but ASCII85 is not.
- Generic processing-stage API, derivation-based provenance, POSIX/platform abstraction, and correlation identifier semantics: not addressed.
- CI remains red for e2e-base and Playwright, with no newer run at this head.

ajroetker:
[codex] I opened #392 with fixes for the concrete review blockers plus the platform/source-fingerprint cleanups. It is based on an exact mirror of this PR's reported head (01c28ba), because the visible fork branch currently points to a partial diverged history. The generic processing-stage and derivation-provenance schema redesign remains separate because it needs a coordinated public API/SDK migration.

ajroetker:
merged

PR #369, opened 2026-07-16
Make Antfarm inference connection-first

timkaye11:
## Summary

- add a Connections API inventory with a synthetic local-inference connection and bounded provider model discovery
- proxy Antfly-compatible inference operations server-side so provider credentials never reach the browser
- make Antfarm model inventory, playgrounds, semantic search, and index forms select explicit connections
- deduplicate concurrent model inventory requests, bound the browser cache, preserve provider model IDs, and clarify unsupported playground actions
- validate inference operation names and require HTTP(S) provider origins

## Production notes

- provider model discovery uses bounded fanout and a short-lived server cache
- invocation operations are allowlisted and provider secrets are resolved only on the server
- local inference remains an explicit connection instead of an implicit fallback

## Validation

- zig build root-test (221 passed)
- zig build lib-api-connections-test (11 passed)
- pnpm --filter antfarm typecheck
- focused Antfarm Vitest suites (9 passed)
- Biome and diff checks


bpopadiuk:
Sol 5.6 Extra High review:

```
Verdict: **request changes**. I found three merge-blocking issues at exact head `5a6bf09d`. Nothing was posted to GitHub.

1. **[P1] Per-connection capabilities are not enforced.**  
   [connections.zig:160](/private/tmp/antfly-pr369-review/zig/pkg/antfly/src/api/connections.zig:160) checks only the global operation allowlist. It never checks the selected connection’s required `capabilities`, so an embed-only connection can still proxy `generate`, `transcribe`, or any other allowlisted operation using its secret. Add an operation-to-capability mapping and reject operations the connection does not declare.

2. **[P1] Secret references are passed literally during model discovery.**  
   The configured API key remains unresolved at [connections.zig:925](/private/tmp/antfly-pr369-review/zig/pkg/antfly/src/api/connections.zig:925), and model listing turns it directly into a Bearer header. Therefore documented values such as `${secret:openai.api_key}` fail authentication. Invocation resolves the same reference correctly, but `/connections?include=models` marks the connection as errored; Antfarm then requires `status === "connected"` at [BackendInfoBar.tsx:27](/private/tmp/antfly-pr369-review/ts/apps/antfarm/src/components/playground/BackendInfoBar.tsx:27) and exposes no usable models. Resolve secrets just-in-time for model discovery without placing resolved values in responses or caches.

3. **[P1] `local-inference` can collide with a configured connection ID and silently misroute traffic.**  
   The inventory unconditionally creates the synthetic ID at [connections.zig:467](/private/tmp/antfly-pr369-review/zig/pkg/antfly/src/api/connections.zig:467), while configuration accepts arbitrary connection keys. The browser then treats every connection named `local-inference` as the synthetic runtime and bypasses the proxy at [api-config-provider.tsx:82](/private/tmp/antfly-pr369-review/ts/apps/antfarm/src/components/api-config-provider.tsx:82). Such a configured connection becomes duplicated or unreachable, and on a node without local inference its requests go to the wrong endpoint. Reserve/reject the ID or use a collision-proof runtime identity instead of routing by a magic string.

Validation: the focused Zig connections suite passed all 11 tests, and TypeScript CI—including Playwright—passed. The exact head is not currently green: Zig `e2e-base` has three failures in apparently unrelated areas, and Go SDK CI failed during checkout before tests ran. Those failures look unrelated to this diff, but both workflows still need a clean rerun before merge.
```

PR #371, opened 2026-07-17
Fix restore, split, and HA lifecycle races

ajroetker:
## Summary

- restore the storage-generation, raft snapshot, metadata incarnation, and split-publication hardening from #364 after its merge was reverted by the subsequent branch merge
- preserve the newer inflight restore, split checkpoint, exact delta sequencing, and HA lifecycle fixes already on this branch
- fence runtime-status refreshes and live publications by table epoch and observation generation, including rejected-publication ownership cleanup
- restore durable split attempt epochs, destination bootstrap checkpoints, source acknowledgements, and snapshot payload publication
 
## Revert history
- PR #364 was merged into an intermediate full-text branch, reverted there, and that reverted topology was later merged to main. This PR selectively reapplies #364's actual diff instead of reverting the broad 258-file merge revert, which would also restore unrelated branch history.

## Linked tickets

Addresses #398
Addresses #399
Addresses #400


zig/pkg/antfly/src/api/e2e.zig
```diff
@@ -461,19 +478,47 @@ test "public api smoke e2e creates table inserts and queries documents" {
...
     );
     var provisioned_write_source = table_writes.ProvisionedTableWriteSource.init(
         replica_root,
         table_catalog.CatalogSource.fromMetadataService(&svc),
     );
+    var provisioned_storage = ProvisionedGroupStorage.init(std.testing.allocator);
+    defer {
+        // Attached sources borrow storage callbacks. Drain their workers before
+        // destroying the callback owner, matching DataServer's teardown order.
+        provisioned_write_source.deinit();
```
bpopadiuk: Codex (GPT-5): [P1] This order does not actually drain the cached DB workers before destroying their query-visibility callback owner. `ProvisionedTableWriteSource.deinit()` leaves the write-cache DB hooks attached and then poisons the source; `ProvisionedGroupStorage.deinit()` closes those DBs afterward, allowing a derived worker to call `onManagedDerivedVisibilityChanged` through the dead pointer. This reproduced as a `0xaaaaaaaaaaaaaaaa` bus error, and production `DataServer.deinit()` uses the same ordering. Detach/drain the DB callbacks while the source is still alive.

bpopadiuk (dismissed):
Codex (GPT-5):

Requesting changes for one confirmed memory-safety/lifecycle blocker on exact head `dec4d72251c1c7d9632da9d6dd84bdfc410f1ac5`.

The new teardown path destroys `ProvisionedTableWriteSource` before `ProvisionedGroupStorage`, but source deinit does not detach or drain the cached DBs' query-visibility callbacks before poisoning itself. Storage teardown can then drain a derived worker that invokes `onManagedDerivedVisibilityChanged` through the dead source. I reproduced this as a bus error at `0xaaaaaaaaaaaaaaaa` in `table_writes.zig:6965`.

This is not limited to the fixture: production `DataServer.deinit()` uses the same write-source-then-storage ordering. Please detach all visibility hooks and drain/close the cached DB workers while the source remains valid, or otherwise change the ownership/teardown contract so no callback can run after source deinit.

Validation on this head:
- `zig build public-api-parity-test`: aborts reproducibly.
- `zig build unit-test`: fails on the same parity artifact.
- `zig build root-test`: 177/177 passed.
- The final schema-reconciliation fix passed the focused migration E2E 5/5.

dovinmu:
```
  ### Blocking findings

  1. Split acknowledgements can permanently wedge replay
     data/storage/shard_state_store.zig:1335, 1543–1544, 1676–1702

     appendOperationEffects reads the durable delta sequence before processing the batch, validates acknowledgements against that stale value, and only emits the batch’s delta afterward. If replay batches a write and its
     later acknowledgement together, the acknowledgement is rejected as SplitAcknowledgementAheadOfSource. Retrying sees the same state and fails forever.

  2. Split delta sequence numbers are replica/batching-dependent
     shard_state_store.zig:1676–1702, data/runtime.zig:5908–5926, 6216–6224

     Only one delta sequence is allocated per applyBatch, not per replicated entry. Different replicas can apply the same log using different batch boundaries and therefore assign different sequence numbers. The PR now raft-
     replicates acknowledgements and uses them as resume cursors, treating those local numbers as cluster-stable. After leader failover this can skip unreplicated boundary writes or permanently wedge catch-up.

  3. Existing metadata range records are no longer decodable
     metadata/storage/raft_apply_store.zig:4320–4346, 4675–4727

     split_attempt_epoch was appended to range encoding, but the decoder unconditionally reads it while the enclosing metadata snapshot format remains version 1. Old snapshots/log records end before this field and fail with
     InvalidMetadataTransitionEncoding. This appears to break upgrades for ordinary existing ranges, not just in-flight splits.

     There is a similar unversioned compatibility break for data split state: currentSplitAcknowledgement now requires exactly 32 bytes at shard_state_store.zig:308–316, while the old acknowledgement was 16 bytes.

  4. Snapshot installation relies on consumed map capacity
     data/storage/raft_apply_store.zig:589–590, 755–759

     Installation reserves capacity in shard.batches, releases the shard mutex during lengthy staging, then uses putAssumeCapacity. Another group mapped to the same shard can consume the reserved capacity meanwhile. The
     final insertion can assert in debug or violate the hash map’s capacity contract in release builds.

  ### Additional actionable findings

  5. Bulk-ingest finish errors clear retry state
     api/table_writes.zig:10334–10356

     Even if one DB’s finish operation fails, the code clears every entry’s bulk-ingest state and removes the active session before returning the error. Retrying then sees no session and returns success, even though
     publication may not have completed.

  6. Error-path search-result leak
     storage/db/query/result_shape.zig:1067–1079

     errdefer filtered.deinit() is installed only after applyStoredSearchPatternFilters. If that call fails, the already-owned filtered result leaks. The vector path around lines 1104–1125 has the analogous ownership gap.

  7. Several test suites stopped gating zig-unit-test
     zig/build.zig:5511–5552

     Root DB, metadata logic/service, storage-maintenance, dense-index-lifecycle, and related suites are no longer dependencies of the normal unit-test step. Their broad replacement is attached to unit-test-progress, which
     CI does not invoke.
```

dovinmu:
[codex with linux VM]

# Antfly PR #371 Linux lifecycle qualification

Date: 2026-07-23/24 UTC  
PR: https://github.com/antflydb/antfly/pull/371  
CI run investigated: https://github.com/antflydb/antfly/actions/runs/30041987082  
Verdict: **merge blocker**

## Subjects and environment

- PR head: `9c54dba3927d5168677d15913bad8fd9bf4b80d9`
- PR binary SHA-256: `3f26e3becb13b02f566e27ab6989db72dc4fd07e03ed42f75d065960585b4fa0`
- Base: `8f063f7bdae349401a47767ec55a4e02277a3592`
- Base binary SHA-256: `375d56f8dc9b6fd2ae3998c938d6a4f051bb997b2619a0fde54c455d45b0cea1`
- Build: Zig 0.16.0, repository CI Debug/full build script, x86_64-linux-gnu baseline CPU
- Host: disposable GCE `n2-standard-8`, Ubuntu 24.04
- Container: Ubuntu 24.04, CPUs `0-7`, 15 GiB hard memory limit, 15 GiB memory+swap limit
- Final cgroup memory events: `oom 0`, `oom_kill 0`, `oom_group_kill 0`

Each repetition used a fresh pytest fixture lifecycle and that subject's own
test harness/CLI flags. This matters because the base still invokes the legacy
`--tick-ms 25` interface while the PR uses the split tick controls.

## Results

| Subject / lane | Result |
|---|---:|
| PR autosplit focused | **99 passed / 1 failed** |
| PR schema migration focused | **98 passed / 2 failed** |
| Base autosplit focused | 6 passed / 4 failed |
| Base autosplit smoke | 0 passed / 1 failed |
| Base schema migration focused + smoke | 11 passed / 0 failed |
| PR full Antfly e2e | 150 passed, 3 skipped, 61 deselected |
| PR full inference e2e | 34 passed, 86 deselected |
| Base full Antfly e2e | 150 passed, 3 skipped, 61 deselected |
| Base full inference e2e | 34 passed, 86 deselected |

The requested stop point was one complete PR/base full-suite pair. A second PR
full suite had emitted only its 11-line startup/collection prefix when stopped;
it is preserved but is not counted.

## Blocker 1: PR autosplit can strand indefinitely

PR focused autosplit repetition 23 failed after 181.95 seconds:

```text
AssertionError: table did not finalize an automatic split
split_observations: []
queued_split_transitions: 1
completed_split_transitions: 0
```

The transition repeatedly logged:

```text
split transition step failed ... phase=bootstrap_peer err=GroupLeaderUnavailable
```

This is not the exact `RuntimeStatusPublicationFenced` signature from the first
CI attempt. It is a second route to the same externally visible result: the
table remains on one shard and the split never publishes/finalizes. Repetition
76 also contained `GroupLeaderUnavailable` but eventually recovered and passed
after roughly 84 seconds, demonstrating a long-tail recovery mode in addition
to the terminal failure.

Impact if shipped: an automatic split can remain queued indefinitely, leaving
the original shard hot/oversized and preventing the expected capacity
redistribution. Operational retry/reallocation does not guarantee recovery
within the test's three-minute window.

## Blocker 2: CI schema-migration 503 reproduced twice

PR focused schema-migration repetitions 55 and 86 independently failed with
the same response reported by CI:

```text
HTTPError: 503 Service Unavailable
body=standby read unavailable
```

Both failures occurred on the post-migration read of `doc-0500`, after logs
reported structural reconciliation complete and queued index repair. The next
fresh repetition passed in each case.

Impact if shipped: after a schema/full-text migration is reported complete,
ordinary document reads can transiently return 503 because the serving view is
still treated as an unavailable standby. Clients would observe downtime or
need retries across a supposedly completed migration.

## Interpretation

The broad full suites passing once per subject do not make these failures
benign. The focused Linux soak reproduced:

- the schema-migration CI failure exactly, twice in 100 runs; and
- an autosplit non-finalization on the exact PR head through a distinct
  `GroupLeaderUnavailable` bootstrap failure.

Both failures immediately disappeared on a fresh lifecycle, which explains why
single CI reruns alternate between green and red. That is evidence of
intermittent lifecycle races, not evidence that the assertions are flaky.

Recommendation: do not merge or include PR #371 in the next release until both
post-transition availability and split-bootstrap retry/fencing behavior are
fixed, then rerun the same 100x focused Linux qualification.

dovinmu:
[opus + codex soak testing]
```
## External QA report

WHAT WAS ACTUALLY RUN: main d4b35bdd9 + PR #371 head 07312188f (which already contained
#401 b021192502 and #393 855117c1a), merged locally as 79d664668a. Every measurement
below comes from that build.

WHAT WAS ONLY READ: the 24 commits since, up to 3bb0ecc48 (118 files, ~19,800 insertions).
Nothing in that range has been executed here. The findings below were re-checked against
it at source level only:
  - the OpenAPI contract mismatch has since been fixed by d44b270e4
  - the readiness check, the models-endpoint timeout, and the chaos-test target are all
    unchanged, so those findings still stand
  - the index-rebuild stall was NOT re-verified; db.zig changed by ~676 lines in that
    range, so treat it as untested against current HEAD

**Two findings should hold 0.2.0. Three others are worth fixing but should not block it.**

---

# Blockers

## 1. Shard splits can strand under mixed load

A table that exceeds its size threshold is supposed to split into two shards. Sometimes the
split is queued and then never completes. The test gives up after 180 seconds.

Frequency depends on what *else* is running, far more than on repetition count:

    idle machine, 100 runs ................. 0 failures
    3 copies of this same test, 501 runs ... 1 failure
    3 *different* test suites, 100 runs .... 2 failures

**If you loop this test on an idle machine it passes every time.** You need concurrent
unrelated work to see it. (A serial soak reporting 100/100 is not evidence the bug is gone.)

### Where it goes wrong

Before a split proceeds, the destination shard must be judged ready. That is one yes/no built
from eight conditions in `transitionStatusHasStablePlacement` (`metadata/service.zig:6439`).
In every failure examined, seven passed and the same one failed: `voter_count_known`.

The leader *is* known and correctly placed, and the replica count *does* match what placement
expects. Only the "is the count known" flag is false.

This reads as a leader problem in the logs and isn't one — `requireGroupReadyForTransition`
(`raft/hosted_shard_ops.zig:379`) reports all eight causes as `GroupLeaderUnavailable`.

Captured live with temporary logging:

    voter_count ambiguity: group=4761281844279739167 existing=4 incoming=3 store=103
    readiness=false group=4761281844279739167 leader_known=true leader_store_id=101
        leader_placed=true voter_count_known=false voter_count=4 expected=4 healthy=4 joint=false

One store reported 3 replicas against an earlier report of 4. That sets an internal ambiguity
flag, which makes `voter_count_known` false, which denies readiness (`metadata/state.zig:736`):

    .voter_count_known = state.observed_voter_count != null and !state.ambiguous_voter_count,
    .voter_count       = if (state.observed_voter_count) |count| count else 0,

Note `voter_count` keeps the *first* value and discards the conflicting one, while the ambiguity
flag remembers the conflict. That is why diagnostic dumps look self-consistent (4 expected,
4 observed) while readiness is still refused — the disagreeing value isn't visible anywhere.

Related: reports carry their own `voter_set_known` flag saying whether the sender knows the
replica set. `state.zig` never reads it; the check at line 647 is only
`if (group_status.voter_count > 0)`. So a replica that says "I don't know" still contributes a
number that can cause the conflict.

### What is NOT proven

**The mechanism is localised; the root cause is not fully established.** We never captured
whether the disagreeing report had `voter_set_known=false` — and that is exactly the fact the
obvious fix depends on. Treat the fix direction below as a hypothesis, not a conclusion.

Also unknown: whether the stranded state ever recovers (the test stops at 180s, so "never
recovers" was never shown), and whether refusing readiness is the bug or a correct response to
something upstream.

### If you attempt the obvious fix

Adding `and group_status.voter_set_known` to line 647 makes the case pass but breaks three
existing tests:

    metadata state conservatively aggregates readiness across healthy peers without leader truth
    metadata state tracks authoritative voter count when healthy peers agree
    metadata state ignores group status from stores without current placement

They assume reports without a known replica set still supply an initial count. A fix must
separate "provide the first value" from "count as a conflict". `metadata/reconciler.zig:2295`
computes the same expression separately and needs the same treatment.

**Don't fix this by adjusting retry timing.** The retry backoff added in #401 was tested
directly and is working: after a backfill, CPU drops to ~1% of one core and stays there; retry
counters never regressed across 65 samples under load; all five retry-reason counters stayed at
zero. Retries are fine — readiness never becomes true.

Also: the existing test for this check (`metadata/service.zig:6527`) hard-codes
`voter_count_known = true`. The one condition that fails in practice is never exercised.

**Cheapest high-value change:** make that error report which of the eight conditions failed
instead of always saying `GroupLeaderUnavailable`. Diagnosing this needed source reading; it
should have needed one log line.

**Release criterion:** either fix it, or demonstrate the split recovers within a defined bound.

---

## 2. Full-text index rebuild stalls under moderate load

`test_schema_migration_full_text_rebuild` changes a table's schema and waits up to 240 seconds
for the index to rebuild.

Same binary, both conditions:

    idle machine, 100 runs ...... 100 passed, median 9.0s (max 11.3s)
    busy machine,  12 runs ......   0 passed, 247.7s - 249.4s every time

All failures: `full_text_index_v1 did not finish rebuilding in time`. A 1.6-second spread
across twelve runs means every one hit the deadline rather than finishing slowly. A 9-second
operation not completing in 240 seconds is a stall.

"Busy" is three concurrent test suites — 1-minute load average ~2.7 on a 14-core machine. Not
saturation.

Two details suggesting this is specific rather than general slowness: during the same period on
the same machine, the shard-split test passed 60/60 and the table-drop stress test 25/25. And
the *first* index rebuild in the test completes fine; the second one hangs.

### Caveat

We ruled out our own instrumentation as the cause (re-ran with an unmodified binary, still
0/8). We did **not** rule out test-process or resource interference from the concurrent
suites — that would need one more discriminating experiment, e.g. reproducing under synthetic
CPU/IO pressure rather than other test processes. Worth doing before concluding it is a
product defect, but the idle-vs-loaded contrast is stark enough that we would not ship without
either fixing it or showing it is a harness artifact.

Not otherwise investigated. Suggested start: make the rebuild's completion check report *why*
it hasn't finished — starved, deadlocked, or waiting on something that never arrives.

---

# Worth fixing, but not release blockers

## 3. `lib-raft-chaos-test` selects zero tests  (one-line fix)

`zig build lib-raft-chaos-test` fails in about a second because all 29 of its `--test-filter`
names match nothing. The tests live in `pkg/antfly/src/raft/sim_harness.zig`; the step compiles
`root.zig`, which imports `metadata/sim_harness.zig` but never the raft one.

`2fc3bde65` added `raft_sim_test_root.zig` and pointed `lib-raft-sim-test` at it (now 10/10).
The chaos step eight lines below at `build.zig:3479` still uses `lib_test_mod`.

**Pre-existing on `main`, and a CI-routing defect rather than a product defect.** It matters
because the dead filters cover queued split/merge transitions, restart resumption, mid-flight
rollback and same-id overwrite races — the same area as blocker 1. It becomes release-relevant
only if you want that lane as a release gate. Fix: point
`lib_raft_chaos_tests.root_module` at `raft_sim_test_mod`, and expect real failures once it
runs.

## 4. `/ai/v1/models` latency blocks a regression test in CI

The endpoint consistently takes ~2.4s (903 bytes, 4 models, warm server, every request;
`/db/v1/status` on the same server answers in under a millisecond). Looks like a per-request
models-directory rescan rather than a cached list.

`e2e/antfly/test_standalone.py::_wait_for_server` polls it with a 2-second timeout and needs
two consecutive successes, so it always gives up — making
`test_standalone_drop_tables_with_pending_embedded_embeddings` fail during setup 100% of the
time (3/3). That test also carries `standalone_integration`, which the default `e2e-base` tier
excludes, so CI never runs it either way.

**Important: this is a timeout mismatch and a CI coverage gap, not evidence of a product
regression.** With the timeout raised, that test passes **120/120** (60 Metal, 60 forced
fallback) with zero crashes. The #388/#391 table-drop crash fix demonstrably holds — nothing in
CI is currently checking that, which is the actual problem.

Fix the readiness probe and get the test into CI. Optimise the endpoint only if you have a
catalog-latency requirement.

## 5. `projection_checkpoint_config_hash` → `..._config_fingerprint`

`3f4194870` renamed this field in `specs/openapi/antfly/indexes.yaml` (6 places) without
updating four assertions in `api/openapi_contract.zig`, so `zig build unit-test` and GitHub
`zig-base` failed. `d44b270e4` has since updated the assertions and CI is green again.

**Not a stable-release break.** `specs/openapi/antfly/indexes.yaml` does not exist in v0.1.2 or
v0.1.3; the field first appears in v0.2.0-rc.20. So this only affects consumers pinned to an
earlier 0.2.0 RC.

Decide the intended name and document it. Only a release concern if you are promising
RC-to-RC compatibility.

---

# Smaller issues

- The standalone server logs its own address as a raw struct:
  `standalone public api listening on http://.{ .ip4 = .{ .bytes = { 127, 0, 0, 1 }, ... } }`
  The health-server line above it is correct. Harnesses parse these lines for URLs.
- `GET /ai/v1/models` lists `antflydb/clipclap` twice of four entries; clients k
[message truncated]
```

PR #375, opened 2026-07-18
Fix Linux musl statvfs builds

bpopadiuk:
Codex (GPT-5):

## Summary

- move the `statvfs` ABI boundary behind a small C shim so Zig does not translate musl's opaque `struct statvfs`
- centralize platform-module construction so native libc, libc-enabled tests, standalone inference, and no-libc/WASM graphs attach the shim coherently
- keep this PR scoped to the release build repair; CI boundary hardening is tracked in #376

## Root cause

Normal CI compiled and ran the Zig suites against the host/glibc target. The tagged release was the path that compiled Linux musl packages. Zig 0.16 translates musl's `struct statvfs` as opaque, so `filesystem.zig` could not instantiate it.

The initial shim exposed a second build-graph edge: `builtin.link_libc` follows the final artifact, while the shim had been attached according to a top-level build option. A libc-enabled test artifact could therefore emit the shim call even when its implementation was omitted. Platform construction is now shared, and libc-enabled artifact roots attach the shim when consuming a platform module configured without libc.

## Impact

This repairs the Linux runtime archive build and unblocks downstream container/release jobs. The failed `v0.2.0-rc.19` release will need a new Antfly tag after this merges.

## Validation

- [x] reviewer regression: `zig build -j2 resource-budget-test -Dlink-libc=false -Dha-tests=false --summary all` — 13/13 steps, capacity 2/2, resource budget 27/27
- [x] normal-libc resource/capacity suite — 13/13 steps, capacity 2/2, resource budget 27/27
- [x] standalone platform package graph
- [x] standalone inference native and WASM package graphs
- [x] linked `x86_64-linux-musl` `ReleaseFast` capacity probe
- [x] linked `aarch64-linux-musl` `ReleaseSmall` capacity probe
- [x] negative control without the C source fails with `undefined symbol: antfly_platform_filesystem_capacity`
- [x] exact x86_64 release-package build produced the runtime archive and C API library
- [x] native C-API test bucket with the failed CI seed — 10/10 passed
- [ ] current PR checks

## Follow-up

#376 covers a real linked PR/merge-group musl boundary gate, complete package-input detection, and cost-aware post-`main` package validation.


.github/workflows/antfly-linux-release-build.yml
```diff
@@ -0,0 +1,90 @@
+name: Linux Release Archives
```
ajroetker: Why did this get it's own ci file now? just curious

ajroetker: i see we're adding another build step for this on prs, we probably shouldn't do that as it'll burn our github actions credits which is why i didn't do it before

.github/workflows/zig-tests.yml
```diff
@@ -139,6 +155,22 @@ jobs:
             echo "trace_validate=true" >> "$GITHUB_OUTPUT"
           fi
 
+  release-linux-compile:
+    name: release-linux-compile
+    needs: changes
+    if: >-
+      needs.changes.outputs.release_targets == 'true' &&
+      (github.event_name == 'pull_request' ||
```
ajroetker: I don't think this should run on prs probably just merges

.github/workflows/antfly-linux-release-build.yml
bpopadiuk: i think the separate ci file is so that it can be reused across preflight checks and release build workflows.

seems like you're right about cost though. maybe this should just run on merge to main? think we need some assurance that main compiles prior to cutting release images, but maybe not each pr. maybe it would make sense to persist an artifact for each merge to main and then have release tags just be a promotion exercise instead of building from scratch

bpopadiuk: think we can build just the translation boundary to catch these, without building an entire release tag. working on setting that up

PR #380, opened 2026-07-21
Route selective dynamic filters through exact vector search

dovinmu:
> ## Summary
> 
> - emit typed doc values for dynamically inferred numeric and boolean fields
> - allow covered inferred fields to participate in native structured-filter planning
> - expose native candidate count and exact-versus-HBC route reasons in dense-search profiles
> - validate exact top-100 results for numeric and categorical 1% filters
> 
> ## Root cause
> 
> The default `infer_types` schema indexed dynamic string fields, so categorical term filters could resolve to native ordinal sets. Dynamic numeric fields did not emit typed doc values, so numeric range predicates could not produce native `filter_ids`. Dense search consequently missed the existing exact-native-filter route and fell through to HBC plus post-filtering.
> 
> This change fixes only that mapper/planner handoff. It does not change HBC.
> 
> ## Impact
> 
> Selective dynamic numeric filters with native coverage can now exact-score candidate sets within the existing budget, improving filtered-vector recall and avoiding unnecessary HBC work. Planning remains fail-closed when typed doc-value coverage is missing, sparse, malformed, or incompatible.
> 
> ## Validation
> 
> - `zig build lib-db-query-test -Doptimize=Debug`
> - 168 passed, 2 skipped, 0 failed, 0 leaked
> - includes 10,000-vector numeric and categorical cases with 100 eligible vectors and exact top-100 assertions
> - includes the 500-candidate exact-route budget decision
>


ajroetker:
## Codex review — request changes

1. **P1 — Dynamic date/geopoint inference is dropped when a document also contains a numeric/boolean field.** [document_mapper.zig:2007](https://github.com/antflydb/antfly/blob/1436a17b1946a793d29beb85d3eaa4e9fe33454b/zig/pkg/antfly/src/storage/db/document_mapper.zig#L2007-L2011) only projects numeric and boolean values. Once one is appended, typed_fields becomes non-null and the mapper suppresses typed_source, preventing the existing inference pass from collecting dates and geopoints. The existing schema-present infer_types test demonstrates this: its date query expected one hit and received zero, causing [zig-base to fail](https://github.com/antflydb/antfly/actions/runs/29806351527/job/88557728583). Preserve the fallback source or project every previously supported inferred type.

2. **P2 — Missing optional exact callback now disables the existing exact fallback.** [search_exec.zig:12287](https://github.com/antflydb/antfly/blob/1436a17b1946a793d29beb85d3eaa4e9fe33454b/zig/pkg/antfly/src/storage/db/query/search_exec.zig#L12287-L12299) changes a qualifying route to HBC whenever exact_dense_search is null. Previously null invoked exactScoreNativeDenseFilter; that fallback remains below but is now unreachable. Any executor relying on the optional callback default loses exact filtered results and gets approximate HBC results instead.

ajroetker:
## Codex follow-up

Commit [4254084](https://github.com/antflydb/antfly/commit/4254084e0bac1ceaa38eda31add08eb1e47eed24) addresses the P1 mapper regression. Dynamic inference now uses detectTypedFieldProjectionValue, preserving inferred dates and geopoints when numeric or boolean projected fields are also present, and the previously failing reopen test has been added to the selected test suite. The new zig-base check is currently in progress.

The P2 exact-route issue remains: [search_exec.zig:12287](https://github.com/antflydb/antfly/blob/4254084e0bac1ceaa38eda31add08eb1e47eed24/zig/pkg/antfly/src/storage/db/query/search_exec.zig#L12287-L12299) still changes a qualifying exact route to HBC whenever exact_dense_search is null. This bypasses the existing exactScoreNativeDenseFilter fallback immediately below it, which was used before this PR. Please remove that route rewrite or otherwise retain the local exact fallback for executors that omit the optional callback.

PR #381, opened 2026-07-21
[bug] Fix unbounded disk growth during large (backfill) enricher transactions

excenter:
## Summary

- scan enricher/backfill ranges in windows: collect one batch, close the pebble iterator, run `ProcessBatch`, resume from the exact cut key (inclusive)
- never hold an iterator open across `ProcessBatch` — in the enricher backfills that callback makes remote embedder calls per batch, so a single scan-long iterator lives for hours
- clamp `BatchSize < 1` to 1: windowing turns that degenerate input from "one-item batches" into a silent infinite loop that never advances the lower bound
- document the semantics change on both functions: the traversal is no longer a single point-in-time snapshot — an item enriched between windows is skipped, not re-embedded (desired for backfills)
- cross-compile the builder stage natively via `--platform=$BUILDPLATFORM` (the Dockerfile already passes `GOOS`/`GOARCH` from `TARGETOS`/`TARGETARCH`; without it the amd64 half of a multi-arch build compiles under emulation)

## Root cause

`ScanForEnrichment` and `ScanForBackfill` invoked `ProcessBatch` from inside the scan callback, holding one pebble iterator open for the entire traversal. An open iterator pins the version it was created against, so sstables obsoleted by concurrent flushes and compactions cannot be deleted until it closes. The enricher backfills call a remote embedder per batch, so the iterator lives for hours and disk grows at write-rate × duration, unbounded — reclaimed only on process restart, when pebble removes orphaned files on `Open`.

On a full volume this deadlocks: reclaim needs the iterator closed, closing needs the scan to finish, and the scan cannot proceed without disk. A crash becomes the only recovery path.

Observed in production (3 clusters, v0.1.2): a ~5GB corpus backfill grew data-node volumes at ~20–25GB/h (~40× write amplification), filling a 180Gi volume in 9 hours. Nodes that crashed mid-backfill "recovered" disk on restart (the orphan sweep on `Open`); nodes that survived pegged at 100% until intervention.

## Fix

Windowed scans. Cut the window at the incoming primary key when the batch is full: every key strictly before the cut has been consumed, so resuming at exactly that key (pebble's inclusive `LowerBound`) neither re-emits nor drops items, even when doc keys are prefixes of one another and their key groups interleave (`user` / `user:1` / `user:10`). Enrichment keys that sort before a not-yet-reached primary stay parked in `seenEnrichments` across windows, same as within a single pass — this is what keeps the pipeline case (embedding `:e` sorts before summary-primary `:s`) correct across a cut.

## Evidence

Results after deploying this patch (built on v0.1.2) to three production clusters:

- the same 105k-document backfill that previously grew each data node by ~190GB overnight now completes with disk oscillating ±1GB — live compaction reclaim during the backfill, previously impossible
- a wedged node at 190GB dropped to 5.2GB the moment headroom let its iterator close
- incremental backfill: +1.5MB total disk movement (previously GB/hour)

Red/green for the `BatchSize` clamp — with the clamp removed, the new prefix-interleave test (which includes `BatchSize: 0`) demonstrates the loop:

```text
panic: test timed out after 15s
        running tests:
                TestScanForEnrichment_WindowCutPrefixInterleavedKeys (15s)
```

With the clamp: `go test ./src/store/storeutils/...` passes, including `-race`, plus all 17 packages that transitively import `storeutils`.

New regressions:

- `TestScanForEnrichment_WindowedIteratorReleased` — proves `ProcessBatch` runs with the iterator closed: a write made during one batch is visible to the next window's fresh iterator (under a scan-long iterator the write would be invisible and the item re-embedded)
- `TestScanForEnrichment_WindowCutPrefixInterleavedKeys` — every doc emitted exactly once across `BatchSize ∈ {0,1,2,3,10}` with prefix-interleaved keys
- `TestScanForBackfill_WindowCutPreservesSummaries` — a document's summaries survive a window cut at its own boundary

## Out of scope (pre-existing)

`ScanForBackfill`'s grouping drops the shorter document's summary when doc keys are prefixes of one another (`user` / `user:1`): `user`'s `:s` key sorts after `user:1`'s entire group, arriving after `currentDoc` has moved on. Reproduced on v0.1.x with a single window (no cut involved) — this PR processes the identical key stream and leaves that behavior unchanged. The same adjacency pattern causes `ScanForEnrichment` to re-embed an already-enriched prefix document. Happy to file separately.

## Review note

~83% of the `scanner.go` hunks are re-indentation from wrapping the callbacks in the window loop — the true delta is ~91 added / 29 removed lines. Best viewed with whitespace hidden.

- made with robots 🤖

## Follow-up hardening on the current head

This supersedes the earlier exact-cut-key wording above for summary-primary scans:

- summary-to-embedding windows resume at the earlier of the primary (`:s`) and enrichment (`:e`) keys, and clear cached boundary state before reopening; insertions and deletions during `ProcessBatch` are now both observed without redundant embedding work
- Antfly now pins Bleve `v2.5.8-antfly004` and Zapx `v17.0.2-antfly006`
- Bleve `v2.5.8-antfly004` adds `MaxMergePlanInputSize`, a plan-wide persisted-input cap that applies to ordinary non-vector full-text segments; Antfly defaults `max_merge_plan_input_size` to 256 MiB and `segments_per_merge_task` to 5
- the plan-wide cap accounts for Bleve retaining completed task outputs until the full plan is introduced, rather than relying on the vector-only `MaxSegmentFileSize` option

Additional regressions cover summary embeddings inserted and deleted exactly between scan windows, non-vector plan-size enforcement, and runtime config parsing.


excenter:
Before/after ant fly data storage graphs: 
Screenshot 2026-07-21 at 11 18 41 AM

Note the recovered disk storage after the PR was applied 

Screenshot 2026-07-21 at 11 18 54 AM

excenter:
Disk graphs after this PR's fix plus the first bleve GC patch (interim state — the consolidated ledger in the final comment below is the settled account):
Screenshot 2026-07-21 at 3 52 45 PM

excenter:
## Final consolidated report — the complete defect ledger

*(Supersedes my earlier running commentary, now removed — the picture kept growing as we dug. This is the settled account: one disease chain across three codebases. Every fix below has an old-fails/new-passes test on a public branch, linked.)*

**The chain:** a deterministic zapx merge-reader bug made certain merges fail forever → the retry storms amplified through garbage-collection wait-path holes and a fork-added parallel-merge deviation into unbounded disk growth → this repo's scanner held a pebble iterator across hours of embedder calls, pinning obsolete sstables → and independently, the arm64 SIMD decode path silently zeroed location data while reporting success.

### This PR — iterator pinned across network I/O
As described in the PR body; complete for this repo: [`193cad77`](https://github.com/antflydb/antfly/pull/381/commits/193cad7707c56646e30a664025947f48a3fe5d38) (windowed scans), [`798333a8`](https://github.com/antflydb/antfly/pull/381/commits/798333a82f970e87abdf6f33ca5c9a2fd7a46f57) (BatchSize clamp + tests), [`4d89d7d5`](https://github.com/antflydb/antfly/pull/381/commits/4d89d7d5d5db18c8b264358cc369770831f7bd0a) (multi-arch build).

### zapx — branch [`excenter/zapx fix/mergemode-stale-loc-counts`](https://github.com/excenter/zapx/tree/fix/mergemode-stale-loc-counts)

**Merge reader inherits stale location counts — the initiator of everything.** In `nextAtOrAfter`, `numLocValues`/`numFieldIDs` are assigned only when a posting has locations; in merge mode a locations-less posting inherits the *previous* posting's counts, and `mergeTermFreqNormLocs` (which infers hasLocs from count>0) then steals later postings' values → `read location values: EOF`, deterministic for any term appearing in both a term-vectored and a plain field. Read-side only — the segment files were byte-perfect all along; every "corrupt segment" in this saga was this reader bug. Fix: 9-line merge-mode counter reset ([`31854faa`](https://github.com/excenter/zapx/commit/31854faa7a501af481e0c0f275dd4055aa50980e)); the regression test builds the interleaved postings list and fails with exactly that EOF pre-fix ([`15790ca9`](https://github.com/excenter/zapx/commit/15790ca9fbd9be6e4c245eedeba42d416a0fb13a)).

**SIMD StreamVByte decode corrupts column tails on arm64.** Both zapx call sites discard `DecodeStreamVByte32Into`'s `(decoded, dataConsumed)` returns — but the deeper problem is that the returns cannot detect the corruption anyway. Sentinel probe on linux/arm64, one 4-value group, 10 data bytes (no full 16-byte load window anywhere), `dst` pre-filled with `0xDEADBEEF`:

```
control=e4 data=00000100000200000003 len(data)=10
decoded=4 consumed=10 dst=[0 0 0 0]
```

Zero-filled output with a full-success claim. Measured on real segments: 20–71% of a segment's location values decode as zeros on arm64, and merges re-encode the loss permanently. Fix ([`40291c64`](https://github.com/excenter/zapx/commit/40291c64cd52ed449661a06ac8c43a1dd7a7787b), helper [`decodeStreamVByteFull`](https://github.com/excenter/zapx/blob/fix/mergemode-stale-loc-counts/streamvbyte_coder.go#L91)): compute the safe group prefix from the control bytes, hand the kernel only groups with a full 16-byte window, scalar-decode the remainder, and redo everything scalar if `consumed` deviates from the plan. Tests: [`f3d8b350`](https://github.com/excenter/zapx/commit/f3d8b3507b36d3cbb74a18e8c088d8ac6ee7bb81) (tail completion across widths/sizes, truncation past the final group's zero padding), plus the fork's own three arm64 test failures (`TestOpen`, both `TestStreamVByte*`) go red→green under NEON with no env vars. SIMD decode keeps a ~3× benchmark advantage over scalar with the fix in place. Two asks belong upstream in go-highway rather than zapx: the no-window kernel path should return `(0,0)` instead of claiming success, and an arm64 CI job would have caught all of this — every failure reproduces with plain `go test` on any arm64 machine.

### bleve/scorch — branch [`excenter/bleve fix/scorch-forced-gc`](https://github.com/excenter/bleve/tree/fix/scorch-forced-gc)

**Obsolete-segment purge never runs under sustained load — three wait paths, one patch shape.** The persister only calls `removeOldData` when idle, and under churn it never idles: the loop-head `continue`, the merger-catch-up nap select, and the introducer-wait select are three parking spots where orphaned segment files accumulate indefinitely (crash-immune — restart does not clean them; worst case ~9 orphans/s from aborted merge plans alone). Fix: forced-purge tickers in every wait path — [`bbef430b`](https://github.com/excenter/bleve/commit/bbef430ba5ba1c6f03d409c10dc5093bb28d7b2b), [`ad8004eb`](https://github.com/excenter/bleve/commit/ad8004eb89e78fd70570ec7fa6b4061ae820c644), [`723458dc`](https://github.com/excenter/bleve/commit/723458dc9939a0c82851bc39b49311317b69f2a5) — each with a behavioral red/green test: [`df146ff1`](https://github.com/excenter/bleve/commit/df146ff1d8be73a907f9ec9400aa0e54cde4dcad), [`8b583362`](https://github.com/excenter/bleve/commit/8b5833627a9905b5789ac2646a2935ec5a18605f), and [`dc802db1`](https://github.com/excenter/bleve/commit/dc802db1feff7136b6af9b8e949e72c7cb56dea7), the last of which vetoes a merge via a wrapper SegmentPlugin and proves the aborted plan's orphan is swept with zero introductions.

**Parallel merge-task execution stages unbounded GC-immune output.** The fork ran a merge plan's tasks concurrently under a WaitGroup; in-flight merge outputs are correctly `ineligibleForRemoval` (they're live), so retry storms × concurrency staged tens of GB/h that no garbage collector may touch. Fix: restore upstream's serial task execution ([`1e9d7881`](https://github.com/excenter/bleve/commit/1e9d78814f06c28b260839ac19f7d5624f0cbbba) — the serialization half; the merge-cap tuning in that same commit is deployment preference and shouldn't be read as part of the correctness fix).

### Design asks (no code — from the incident evidence)

1. **Merge-failure backoff / quarantine.** A deterministic merge failure currently re-plans and retries in a tight loop forever; that amplifier is what turned a 9-line reader bug into disk exhaustion. Backoff plus a quarantine for repeatedly-failing plans would have capped the blast radius at "one noisy log line."
2. **Parallel embedding-WAL replay.** The post-restart drain is a single goroutine by design — one core replays the entire WAL while the others idle. SIMD restoration made each operation fast; only a design change parallelizes the drain.
3. **Allocation pooling in the generic vector fallback.** Anyone landing on the scalar path pays ~16KB across ~1,000 allocations per 1024-dim distance call (~430× slower than the NEON kernel end-to-end, most of it allocator/GC pressure).

### Status

All of the above are running together in production: zero merge failures since the final fix landed, disks flat with live compaction reclaim during full backfills, SIMD restored end-to-end (NEON frames verified in production CPU profiles on both the query and index-insert paths), and an involuntary node reschedule that recovered in under a minute — the same event previously cost ~an hour of single-core scalar WAL replay. The zapx and bleve branches are structured for direct PRs against your forks; happy to open them whenever you want.

dovinmu:
Hey @excenter, thanks for the PR! This is looking good so far in tests. I'm going to get CI running on this and might have to briefly close-reopen the PR to do it

PR #387, opened 2026-07-22
Implement secret and remote-content reload lifecycle

ajroetker:
## Summary

- add ref-counted, generation/hash-tagged remote-content configuration snapshots with atomic last-known-good reload
- harden secret file change detection for projected-volume inode swaps while preserving request-time secret resolution
- expose non-secret runtime configuration and secret generations, and add an operator config-hash rollout fallback without reading Secret values

## Testing

- `zig build lib-common-secrets-test lib-template-test lib-standalone-runtime-test --summary all`
- `zig build lib-data-runtime-test`
- `env GOWORK=off GOCACHE=/tmp/antfly-go-build-cache go test ./controllers/antfly -count=1`
- `make zig-openapi-generate`
- `git diff --check`

## Follow-up

Colony sources are not part of this repository. This PR exposes and documents the status contract Colony must poll, but the Colony-side wait loop and its integration test must land in the Colony repository.


dovinmu:
[deepseek review]

 PR #387 Review: Secret and Remote-Content Reload Lifecycle

 ### Overall Assessment

 This is a well-designed, thorough PR implementing proper reload lifecycle management for both secrets and
 remote-content configuration. The architecture is sound: ref-counted snapshots for safe concurrent reads, LKG
 (last-known-good) semantics on failure, rate-limited file watching, and clear separation between
 hot-reloadable fields (remote_content) and restart-required fields (everything else).

 I found no existing reviews on this PR. Below are the issues I identified, ordered by severity.

 ────────────────────────────────────────────────────────────────────────────────

 ### Issues That Should Block a Release

 #### 1. IO overhead in statFileMetadata / readFileImage — creates std.Io.Threaded on every call

 Both functions create and destroy a std.Io.Threaded on every invocation. For statFileMetadata, this happens:
 - On every request (via refreshIfChangedForRequest → refreshIfChangedOwned(force_content_check=false))
 - On every health/status poll (via health() → refreshIfChanged())
 - On every secret lookup (via getOwned → refreshIfChangedLocked → statFileMetadata)

 While the request path is rate-limited to 1/s, health() calls are not rate-limited and each forces a full
 readFileImage (file read + SHA-256) if the metadata check passes. Colony's polling loop hitting /status
 frequently would cause redundant file reads and hash computations.

 Recommendation: Add a lightweight rate-limit to refreshIfChanged() (the force_content_check=true path), or
 cache the metadata check result for a short interval (e.g., 500ms). The 1-second throttle already exists on
 the request path; the health path should get similar treatment to avoid thrashing under frequent polling.

 #### 2. Config hash matching between operator annotation and runtime API is untested end-to-end

 The PR's "Follow-up" section acknowledges this:

 │ Colony sources are not part of this repository. This PR exposes and documents the status contract Colony
 │ must poll, but the Colony-side wait loop and its integration test must land in the Colony repository.

 The acknowledgment mechanism depends on:
 - The operator generating completeConfig (via generateCompleteConfig) and hashing it
 - The ConfigMap being synced to the pod's filesystem by Kubernetes exactly as generated
 - The runtime reading the synced file bytes and hashing them identically

 Any discrepancy (trailing newlines, encoding differences, Kubernetes ConfigMap mounting quirks) would cause
 the hashes to mismatch, breaking Colony's wait loop. The E2E tests only test individual components.

 Recommendation: At minimum, add an integration test in this repo that writes a generated config, deploys it
 to a pod (or simulates the ConfigMap sync), and verifies that runtime_config.hash matches the expected
 annotation hash. The Colony integration test should be a blocker for release, not just follow-up. If the
 hashes don't match, Colony will never acknowledge a rollout as complete.

 #### 3. staticConfigHash JSON re-encoding stability is unverified for edge cases

 The function parses JSON → removes remote_content → re-encodes with std.json.fmt → hashes. While this
 pipeline is consistent between startup and reload (both use the same code path), there's no test verifying
 that the encoding is stable across different input formatting. For example:

 ```json
   {"a":1,"remote_content":{},"b":2}
 ```

 vs.

 ```json
   {"b":2,"remote_content":{},"a":1}
 ```

 Both should produce the same hash (they differ only in key order for non-remote_content fields). The
 std.json.parseFromSlice normalizes this. But what about:

 ```json
   {"a":"1.0"}
 ```

 vs.

 ```json
   {"a":"1"}
 ```

 These are semantically different JSON and should produce different hashes (and do, because they're strings).
 But if a human edits the config and accidentally adds/removes insignificant whitespace in numeric fields, the
 hash won't change — which is actually correct behavior. The main risk is if std.json.fmt produces different
 output from what was read, e.g., for floating-point numbers or Unicode escapes.

 Recommendation: Add a test that verifies the hash is stable across whitespace/key-ordering variations of the
 same logical non-remote_content config. The existing test "rejects startup-only replacements" partially
 covers this but doesn't exercise encoding normalization.

 #### 4. RuntimeConfigStatus.stale does not distinguish "no file" from "stale after failure"

 In health():

 ```zig
   .stale_snapshot = self.last_reload_failed,
 ```

 But in the secrets FileStore:

 ```zig
   .stale_snapshot = self.last_reload_failed and self.observed_metadata != null,
 ```

 The secret store correctly only reports stale when there was a previously-observed file that has since become
 unreadable. The remote content runtime reports stale even when the file never existed (if last_reload_failed
 is set by a failed init, though in practice init won't set this flag).

 Recommendation: Align the semantics with the secret store — only report stale when observed_metadata != null.
 Or, if you consider the startup snapshot as "last known good" even without a persistent file, document this
 distinction explicitly.

 ────────────────────────────────────────────────────────────────────────────────

 ### Non-Blocking Issues & Observations

 #### 5. PublishSnapshot.refs uses wrapping add (+%) for retain but checked subtraction for release

 ```zig
   fn retain(self: *PublishedSnapshot) void {
       _ = self.refs.fetchAdd(1, .acq_rel);  // wrapping
   }
   fn release(self: *PublishedSnapshot) void {
       if (self.refs.fetchSub(1, .acq_rel) != 1) return;  // non-wrapping
   }
 ```

 The retain uses fetchAdd which wraps on overflow. For usize (64-bit), this is practically impossible to
 overflow, but a malicious or buggy caller could cause a double-release by overflowing and wrapping back to 1.
 This is extremely unlikely but worth noting.

ajroetker:
Companion Colony acknowledgement PR: antflydb/colony#401. Merge/deploy Antfly runtime status support before enabling Colony's acknowledgement gate.

dovinmu:
[codex]
```
• ## Blocking finding

  - runtime_config.stale can remain permanently true after a transient read failure.
    In remote_content_runtime.zig:174-177

    (https://github.com/antflydb/antfly/blob/bd8426c17c1ef9618f69d860e5520c2ef48e8de8/zig/pkg/antfly/src/common/remote_content_runtime.zig#L174-L177),
    an image matching observed_metadata and observed_hash returns without clearing last_reload_failed. Read/stat
    failures mark the runtime stale but leave the observed identity pointing at the last successfully applied
    file. If the transient error clears and the unchanged file becomes readable again, every subsequent refresh
    takes this early return, so /status continues reporting runtime_config.stale: true indefinitely. That can
    permanently block the acknowledgement loop described by the PR. Recovery should clear the failure when the
    readable image matches the currently published snapshot, while retaining stale state when it merely matches
    a previously rejected candidate.

• DeepSeek did not identify additional release blockers.

  - I/O overhead: real performance concern, but bounded on request paths and mostly pre-existing in the secret
    store. Worth profiling; not inherently blocking.

  - Missing end-to-end hash test: valuable coverage, but Kubernetes preserves ConfigMap data bytes, and both
    hashes derive from the same generated string. Not a demonstrated defect.

  - staticConfigHash stability: object key order may actually produce different hashes, but that only causes a
    conservative restart/stale result—not unsafe hot publication.

  - Stale-with-no-file: not reachable; Runtime.init requires a successfully read file, so any later failure
    necessarily has an LKG snapshot.

  - Refcount overflow: theoretical, non-actionable.

  My previously reported issue remains the only blocker I see: after a transient read/stat failure, recovery to
  the unchanged currently published file does not clear last_reload_failed, potentially leaving /status
  permanently stale and blocking reconciliation.
```

PR #396, opened 2026-07-24
Migrate v0.1.x Go CI to ARC

dovinmu:
## Summary

- run the legacy Go unit job on `arc-antfly-standard`
- run E2E on `arc-antfly-heavy` after unit tests pass
- place large Go, Hugging Face, and Termite caches on the heavy runner's `/mnt/cache` volume
- install PostgreSQL 18 natively instead of relying on Docker-in-Docker

## Why

The restored v0.1.x Go workflow still requested the retired per-run CodeBuild labels, leaving every job queued without a matching runner. ARC is now the active CI runner infrastructure, and its Kubernetes execution model does not support the workflow's old Docker-based PostgreSQL setup.

This backports the existing ARC migration from commit `6f3b7f714` to the v0.1.x maintenance branch.

## Validation

- `make tidy-check`
- `git diff --check origin/v0.1.x...HEAD`
- `actionlint .github/workflows/antfly-go.yml` parses the workflow; its findings are the expected unknown custom ARC labels and pre-existing shell-quoting notices

After this merges, the stale CodeBuild runs can be cancelled and #381 retriggered against ARC.

### DIFF
--- a/.github/workflows/antfly-go.yml
+++ b/.github/workflows/antfly-go.yml
@@ -40,7 +40,7 @@ concurrency:
 
 jobs:
   unit:
-    runs-on: codebuild-GitHubActionsRunnerAntflyAntflyDB-${{ github.run_id }}-${{ github.run_attempt }}
+    runs-on: arc-antfly-standard
     steps:
       - uses: actions/checkout@v6
         with:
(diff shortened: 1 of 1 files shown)


ajroetker:
tests are broken but it ran so that's good! gonna merge

dovinmu:
@ajroetker yeah just making sim not run when that dir was not modified, we can sim soak before 0.1.3. I'll merge when e2e are done

PR #397, opened 2026-07-24
Tune Bleve merge planning for full-text indexes

ajroetker:
## Summary
- configure Antfly full-text indexes with a 256MiB maximum planned vector-segment input footprint and at most 5 segments per merge task
- expose max_segment_file_size and segments_per_merge_task as per-index configuration overrides
- apply the same Bleve scorchMergePlanOptions when creating, opening, and mapping-reloading an index
- validate invalid limits, normalize omitted legacy fields to the Antfly defaults, and log the active values when opening disk indexes

## Why
Serial merge-task execution bounds a merge plan to one in-flight task at a time, but a single task can still stage a large amount of transient output. Bleve already supports these planner options; Antfly previously used bleve.New and bleve.Open, so it never supplied them.

The 256MiB value is a bound on the summed persisted input size used by Bleve when planning vector-bearing merge tasks. It is deliberately not described as a hard output-size cap because actual merge output may differ from the planner estimate.

Keeping these defaults in Antfly avoids changing Bleve library defaults for other consumers while still allowing per-index tuning as production data evolves.

## Compatibility
Existing full-text configurations omit the new fields. Zero values normalize to 256MiB and 5, and explicit default values compare equal to omitted values, avoiding false configuration changes during metadata reconciliation.

## Validation
- env GOWORK=off GOEXPERIMENT=simd go test ./src/store/db/indexes
- env GOWORK=off GOEXPERIMENT=simd go test -run ^$ ./src/store/db/...
- env GOWORK=off GOEXPERIMENT=simd go mod tidy -diff
- go generate ./src/store/db/indexes
- git diff --check

Related incident: https://github.com/antflydb/antfly/pull/381
Related Bleve serialization fix: https://github.com/antflydb/bleve/pull/2


ajroetker:
Superseded by #381. Commit 87b2e91d6efcbf27595d1e0b037a84387c7b187c is now contained in #381 via merge commit 8f0ae65a549cf51eee429564c4c0f2b92bd1578f.

PR #401, opened 2026-07-24
Back off recoverable derived-index retries

bpopadiuk:
Codex (GPT-5): This fixes a derived-index worker retry loop that can consume an entire CPU core after productive import and index work has finished.

## What changed

- apply a bounded 10–250 ms exponential backoff to every recoverable derived-worker failure in both async and io-threaded runtimes
- rate-limit retry logs while including the worker name, concrete error, failure count, and retry delay
- expose derived-worker counts, replay debt, maximum replay lag, and retry counters labeled by failure reason
- expose process user/system CPU time so operators and customers can graph Antfly CPU consumption directly
- serve cached indexing and text-merge snapshots when live write-cache locks are busy, avoiding misleading zero-valued metrics

## Root cause

The previous recovery path only backed off `ResourceBudgetExceeded`. Other recoverable errors, including `WriterLocked`, reset the backoff and immediately yielded before retrying. A persistent recoverable condition could therefore turn into a tight worker loop with no useful progress and very little logging.

A large initial import can activate this path after the import, merge, and compaction work itself has completed.

## Impact

Persistent recoverable failures no longer hot-spin. The added reason-specific counters and replay-debt gauges make the active failure mode and remaining index work visible without requiring a live profiler.

## Validation

- `zig build db-test -- --test-filter 'writer-locked derived-worker retries are counted and backoffed'`
- `zig build db-test -- --test-filter 'io threaded'`
- `zig build api-table-writes-docid-test -- --test-filter 'metrics serve cached snapshot while write cache lock is busy'`
- `zig build lib-data-runtime-test -- --test-filter 'data runtime health metrics include replay debt and provisioned warmup counters'`
- `zig fmt --check` on all changed Zig files
- `git diff --check`

### DIFF
--- a/zig/build.zig
+++ b/zig/build.zig
@@ -4776,6 +4776,7 @@ pub fn build(b: *std.Build) void {
             "provisioned table write source consistent visibility refreshes stale dense status",
             "provisioned table write source visibility hook defers status sampling to runtime owner",
             "provisioned table write source status visibility does not invalidate read cache",
+            "provisioned table write source metrics serve cached snapshot while write cache lock is busy",
             "provisioned table restore lifecycle reserves forwarded owner and caller sources",
             "provisioned startup catch-up enters through forwarded write owner",
             "provisioned runtime status inspection remains available during structural transition",
--- a/zig/lib/platform/src/process_memory.zig
+++ b/zig/lib/platform/src/process_memory.zig
@@ -16,6 +16,9 @@ const std = @import("std");
 const builtin = @import("builtin");
 
 pub const Stats = struct {
+    cpu_available: bool = false,
+    user_cpu_ns: u64 = 0,
+    system_cpu_ns: u64 = 0,
     available: bool = false,
     resident_bytes: u64 = 0,
     peak_resident_bytes: u64 = 0,
@@ -45,6 +48,10 @@ pub fn pressureWorkingSetBytes(stats: Stats) u64 {
 
 pub fn snapshot() Stats {
     var stats = pressureSnapshot();
+    const cpu = processCpuSnapshot();
+    stats.cpu_available = cpu.available;
+    stats.user_cpu_ns = cpu.user_ns;
+    stats.system_cpu_ns = cpu.system_ns;
     if (!stats.available or builtin.os.tag != .macos) return stats;
 
     const malloc_stats = darwin.mallocStats();
@@ -52,6 +59,31 @@ pub fn snapshot() Stats {
     stats.malloc_allocated_bytes = malloc_stats.allocated_bytes;
     stats.malloc_zone_bytes = malloc_stats.zone_bytes;
     return stats;
+}
+
+const CpuStats = struct {
+    available: bool = false,
+    user_ns: u64 = 0,
+    system_ns: u64 = 0,
+};
+
+fn processCpuSnapshot() CpuStats {
+    if (builtin.os.tag == .freestanding or builtin.os.tag == .windows or builtin.os.tag == .wasi) return .{};
+    const usage = std.posix.getrusage(std.posix.rusage.SELF);
+    return .{
+        .available = true,
+        .user_ns = timevalNs(usage.utime),
+        .system_ns = timevalNs(usage.stime),
+    };
+}
+
+fn timevalNs(value: anytype) u64 {
+    if (value.sec < 0 or value.usec < 0) return 0;
+    const seconds: u64 = @intCast(value.sec);
+    const microseconds: u64 = @intCast(value.usec);
+    const seconds_ns = std.math.mul(u64, seconds, std.time.ns_per_s) catch std.math.maxInt(u64);
+    const microseconds_ns = std.math.mul(u64, microseconds, std.time.ns_per_us) catch std.math.maxInt(u64);
+    return seconds_ns +| microseconds_ns;
 }
 
 /// Low-overhead OS memory-pressure sample for hot qualification loops. This
--- a/zig/pkg/antfly/src/api/runtime_status.zig
+++ b/zig/pkg/antfly/src/api/runtime_status.zig
@@ -156,6 +156,7 @@ pub const TableRuntimeSummary = struct {
     indexes_with_replay_debt: usize = 0,
     outstanding_replay_sequences: u64 = 0,
     max_index_replay_backlog: u64 = 0,
+    text_merge: db_mod.types.TextMergeStats = .{},
     async_indexing: db_mod.types.AsyncIndexingStats = .{},
 };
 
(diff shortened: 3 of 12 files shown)


bpopadiuk:
Codex (GPT-5): Reviewed `cce3fb565572c8fac2999318007f3038bc659494` against base `d4b35bdd90c60811a1252fa9ced2992bae876a36`.

No blocking findings.

I traced the retry/backoff paths in both derived runtimes, the counter and replay-debt aggregation, cached metrics fallback, process CPU reporting, and relevant lock ordering. Focused validation passed:

- `zig build db-test -- --test-filter 'writer-locked derived-worker retries are counted and backoffed'`
- `zig build db-test -- --test-filter 'io threaded'`
- `zig build api-table-writes-docid-test -- --test-filter 'metrics serve cached snapshot while write cache lock is busy'`
- `zig build lib-data-runtime-test -- --test-filter 'data runtime health metrics include replay debt and provisioned warmup counters'`
- `zig fmt --check` on all changed Zig files
- `git diff --check`

`zig-base` is green. I also inspected the failing `e2e-base` job:

- The managed-embeddings status assertion reproduces on the exact base commit as well as this head (`replay_applied_sequence < replay_target_sequence` is false in both), so it predates this PR.
- The shard-split failure is stuck in metadata transition handling (`awaiting_split_source_start`) and contains none of the recoverable derived-worker retry conditions changed here; it is unrelated to this diff.

ajroetker:
[codex] Review findings for cce3fb565572c8fac2999318007f3038bc659494:

1. [P2] Replay truncation can still hot-retry without backoff. Both derived runtimes retain a WriterLocked branch that immediately yields with no delay, log, or retry counter:
   - zig/pkg/antfly/src/storage/db/derived/io_threaded_runtime.zig:991-1000
   - zig/pkg/antfly/src/storage/db/derived/async_runtime.zig:638-646

The truncation callback opens a write transaction, so WriterLocked is a real outcome. This leaves one recoverable derived-worker failure outside the stated every-recoverable-failure guarantee and can leave replay truncation deferred until another sequence advances. Please route this branch through the shared backoff/counter path and reset the backoff only after truncation succeeds.

2. [P2] The new Prometheus retry counters can move backward between scrapes. When local_db_mutex is busy, asyncIndexingStatsBestEffort returns an older runtime-status snapshot (table_writes.zig:8140-8144); normal scrapes use current runtime-local counters. Thus antfly_async_index_recoverable_retries_total can fall from a live value to a cached value and then rise again. Writer eviction/reopen can also reset an aggregate component. Prometheus will treat these drops as resets and may produce false retry spikes or alerts. Please keep these counters at stable process scope, or expose snapshot values as gauges with freshness rather than counters.

The principal catch-up and publish paths otherwise look correct and efficient: sleeps occur outside runtime locks, backoff is bounded, logging is rate-limited, and the added atomics are negligible relative to the delay.

Validation on the exact head:
- backoff unit test passed
- IO-threaded runtime tests: 4/4 passed
- cached metrics tests: 7/7 passed
- health metrics tests: 15/15 passed

bpopadiuk:
Codex (GPT-5): I found both review findings valid and addressed them in `d79ff26c5051942b8370accad522a30fd2b506d1`.

1. Replay truncation now retries `WriterLocked` in place through the same bounded 10–250 ms backoff, rate-limited logging, and retry-counter path in both async and io-threaded runtimes. The backoff resets only after truncation succeeds. I applied the same behavior to idle-persist truncation, and the retry loop observes worker shutdown so teardown cannot hang.

2. `antfly_async_index_recoverable_retries_total` remains a Prometheus counter, but its exported values now come from long-lived provisioned-storage counters owned by the shared `ResourceManager`, rather than evictable per-DB runtime snapshots. Live/cache fallback and writer eviction/reopen therefore cannot make the exported series move backward. Worker-count and replay-debt gauges still use the best-effort live/cached snapshot.

Validation:
- focused storage/runtime regressions: 13 passed, 0 failed, 0 leaked
- complete data-runtime test bucket: 70 passed, 0 failed, 0 leaked
- `zig fmt` and `git diff --check` passed

ajroetker:
[codex] Re-review: I found one new blocking issue; I would not approve yet.

- **[P1] The shared async truncation callback now assumes the wrong context type.** [`testRuntimeTruncate`](https://github.com/antflydb/antfly/blob/d79ff26c5051942b8370accad522a30fd2b506d1/zig/pkg/antfly/src/storage/db/derived/async_runtime.zig#L903-L908) casts every callback context to `*TestRuntimeCapture`, but it is reused by tests supplying other structures—for example [`DenseTargetAdvanceSessionCapture`](https://github.com/antflydb/antfly/blob/d79ff26c5051942b8370accad522a30fd2b506d1/zig/pkg/antfly/src/storage/db/derived/async_runtime.zig#L1188-L1226). That is invalid memory access and causes the current Linux `zig-base` job to panic with “incorrect alignment.” On macOS the test passes by layout luck but emits a spurious `WriterLocked`, confirming it is reading unrelated memory. Restore the generic callback as a no-op and add a dedicated capture-aware callback for the new truncation retry test. [Failing CI job](https://github.com/antflydb/antfly/actions/runs/30134165532/job/89614753258).

The two issues from my previous review otherwise appear properly fixed:

- Truncation now uses bounded, shutdown-aware retry backoff.
- Retry metrics now live in `ResourceManager`, avoiding regression when database snapshots are replaced or evicted.

The new focused truncation tests and health-metrics tests passed locally, and `git diff --check` is clean.

bpopadiuk:
Codex (GPT-5): Confirmed—this review was against `d79ff26c5051942b8370accad522a30fd2b506d1` and identified the same test-context corruption behind the linked Linux `zig-base` failure. It is addressed on the current rebased head in `b021192502e5a131d0fd36cdad777a374ed3a9bc`.

I made the callback/context pairing type-safe by giving `DenseTargetAdvanceSessionCapture` its own `testDenseTargetAdvanceSessionTruncate` callback. That callback casts to the correct capture type, records the truncation call and sequence, and the test now asserts that truncation reached sequence 1. I audited the remaining `testRuntimeTruncate` call sites; they all use `TestRuntimeCapture`, preserving the capture-aware `WriterLocked` retry assertions without an invalid cast.

Validation on the PR #371-based tree:

- the previously failing dense target-advance test passes cleanly, without the spurious `WriterLocked`
- async and IO-threaded truncation retry regressions pass
- the shared retry counter/backoff regression passes
- focused storage/runtime set: 4 passed, 0 failed, 0 leaked
- data-runtime health metrics test passes
- write-cache metrics fallback test passes
- `git diff --check` passes

PR #401 is currently clean and mergeable against PR #371's exact head.

PR #404, opened 2026-07-25
inference: make the 0.2.0 model support surface true

dovinmu:
> ## Summary
> 
> This PR makes Antfly's local-inference surface fail closed and match what the current runtime can actually execute. It fixes several GGUF loading and generation defects, replaces the interim release-tier/model-name policy with artifact-derived compatibility checks, adds process-wide resource admission, and documents certification separately from runtime compatibility.
> 
> The runtime now answers two independent questions before loading or running a local model:
> 
> 1. **Compatibility:** can this artifact contract run on an available backend in this build?
> 2. **Resource fit:** can its resident weights and this request's KV/scratch allocation fit safely alongside other admitted work?
> 
> ## Why
> 
> Testing the models implied by the site and API exposed several distinct failure classes:
> 
> - decoder families recognized by name but lacking a usable tensor mapping or forward path;
> - GGUF conversions with unsupported quant formats, missing normalized tensors, or incomplete sidecar bundles;
> - model paths that could panic or exhaust the server rather than return a request error;
> - models that loaded successfully but produced empty, repetitive, or otherwise unusable output;
> - per-request memory checks that did not account for resident models or concurrent requests;
> - public `supported` labels that mixed runtime compatibility with release certification.
> 
> A repository-name allowlist cannot reliably catch those conditions because compatibility depends on the actual architecture metadata, tensors, quantization, sidecars, selected backend, and build configuration.
> 
> ## What changed
> 
> ### Artifact-derived compatibility
> 
> The old `supported | experimental | unsupported` release policy and `support_reason` listing fields are replaced by:
> 
> | State | Meaning | Behavior |
> |---|---|---|
> | `compatible` | Static inspection found a viable runtime contract in this build. | May load, subject to resource admission. |
> | `unknown` | Antfly cannot prove compatibility or incompatibility for the artifact contract. | Rejected unless the server starts with `--allow-unknown-models`. |
> | `incompatible` | The contract is incomplete, unsupported, unsafe, or has no viable backend. | Always rejected; the unknown-model opt-in cannot override it. |
> 
> Compatibility derives from the manifest and artifacts rather than repository IDs. Checks include:
> 
> - bundle and required-sidecar completeness;
> - architecture/runtime availability;
> - GGUF metadata readability, tensor encodings, normalized required tensor names, and backend-specific quant support;
> - SafeTensors shard presence and signatures;
> - ONNX parse/conversion/topology validation when the built-in importer is the available path;
> - whether any backend compiled into the current build accepts the artifact format.
> 
> Inspection results are cached by artifact path plus file size/mtime signatures, including referenced SafeTensors shards, so model listing does not repeatedly inspect large artifacts.
> 
> `/ai/v1/models` now exposes only `compatibility`; certification remains a documentation concern.
> 
> ### Strict explicit backends, safe automatic fallback
> 
> An explicitly selected backend is now strict and never silently changes execution backends. Automatic selection can still fall back when another backend is compatible—for example, from Metal to native when a GGUF quant is supported natively but not release-safe on Metal.
> 
> ### Process-wide resource admission
> 
> A shared reservation ledger now accounts for:
> 
> - resident host and accelerator weights;
> - concurrent host/accelerator KV cache;
> - concurrent host/accelerator scratch memory.
> 
> Model weights are estimated and reserved before backend allocation. ONNX import accounts for both encoded model bytes and converted parameter storage; sharded SafeTensors use the total referenced shard size. Generation requests reserve their KV and scratch estimates against the same ledger and release them with scoped leases.
> 
> Policy limits are stable for a machine configuration. A separate one-shot live-memory-pressure check runs immediately before allocation, avoiding compatibility or policy decisions that flap with momentary system pressure.
> 
> Failures are distinguished as:
> 
> - `MODEL_RESOURCE_LIMIT` (HTTP 400): the model/request cannot fit the configured policy;
> - `MODEL_RESOURCE_BUSY` (HTTP 503, retryable): it could fit alone, but admitted work or current pressure temporarily prevents it.
> 
> ### GGUF loading and generation correctness
> 
> The branch also fixes the underlying defects found during the model matrix investigation, including:
> 
> - preserving structural GGUF facts when overlaying Hugging Face config;
> - Gemma output-scale fallback and Qwen3 tensor-name normalization;
> - byte-level detokenization for `Sequence[Split, ByteLevel]` tokenizers;
> - preserving actionable backend/load errors instead of collapsing them to `NoBackendAvailable` or `NoModelFileFound`;
> - recognizing the real `deepseek4` GGUF architecture spelling, I32 routing tensors, metadata fields, and tensor-name variants;
> - keeping mechanical GGUF tensor mapping independent from serving compatibility policy;
> - projecting Gemma's token-level `<channel|>` boundary so non-streaming and streaming responses expose the final channel while safely withholding reasoning, with fallback when no boundary appears.
> 
> DeepSeek V4 now resolves its real GGUF tensor contract far enough to expose the remaining grouped-output-projection defect, but stays hard-blocked because that forward path is not safe.
> 
> ### Known unsafe/unusable paths remain blocked
> 
> Known process-killing or unusable paths—including standalone CLIP image inference, unsafe ONNX seq2seq/TrOCR paths, incomplete bundles, unsupported hybrid decoder layouts, and generators measured to emit unusable output—remain `incompatible`, even with `--allow-unknown-models`.
> 
> ## Documentation certification
> 
> `docs/guides/supported-models.mdx` now separates runtime compatibility from a docs-maintained certification matrix. It records thoroughly tested artifacts for dense/sparse embedding, reranking, ClipClap, chunking, GLiNER2 extraction, Florence-2 reading, Whisper transcription, and Gemma/Llama/Qwen generation, plus known incompatible models and license caveats.
> 
> In particular, `Babelscape/rebel-large` remains documented as CC-BY-NC-SA (non-commercial) and is not suitable for a default commercial bundle.
> 
> ## Validation
> 
> Local validation on this branch:
> 
> - compatibility-focused inference tests: **27 passed**;
> - shared admission tests: **28 passed**;
> - inference **ReleaseSafe build passed**;
> - full inference suite: **2,066 selected; 2,058 passed; 6 skipped**.
> 
> The full local run emits two existing ClipClap failures: one missing external ONNX fixture and one Metal planner SDPA expectation. The custom test-runner footer reports three failures, although only two failures are emitted and the selected/passed/skipped totals also leave two.
> 
> ## Behavioral changes for review
> 
> - Unknown artifact contracts are rejected by default; use `--allow-unknown-models` only for deliberate experimentation.
> - Known-incompatible contracts cannot be overridden.
> - Explicit backend requests no longer fall back silently.
> - Loading and generation can now return stable resource-limit or temporary-capacity errors before allocation.
> - Public model listings use `compatibility` and no longer expose `support_reason`.
> - “Thoroughly tested/certified” is intentionally maintained in docs rather than encoded as a runtime allowlist.
>


dovinmu:
looking at a fix for this now
```
Blocker: resident-model admission can become permanently exhausted, but the API reports it as retryable contention.

The new process-wide admission accounting keeps weight, tokenizer, and component leases for each LoadedModel. Those leases are released from LoadedModel.deinit(), but loaded models remain in [ModelManager.loaded](https://github.com/antflydb/antfly/blob/87b9322ca1b1f29cda6ada524c09d342b3fead66/zig/pkg/inference/src/server/model_manager.zig#L2215-L2235) until [ModelManager.deinit() at node shutdown](https://github.com/antflydb/antfly/blob/87b9322ca1b1f29cda6ada524c09d342b3fead66/zig/pkg/inference/src/server/model_manager.zig#L2794-L2811). I could not find a loaded-model remove/evict/unload path.

The node config exposes [keep_alive_ms and max_loaded_models](https://github.com/antflydb/antfly/blob/87b9322ca1b1f29cda6ada524c09d342b3fead66/zig/pkg/inference/src/server/server.zig#L221-L239), and the generated API docs say idle/LRU models are unloaded, but these settings are only parsed; they are not used to evict anything. Consequently, after enough distinct models or backend variants fill an admission domain, a later load can return [503 MODEL_RESOURCE_BUSY](https://github.com/antflydb/antfly/blob/87b9322ca1b1f29cda6ada524c09d342b3fead66/zig/pkg/inference/src/server/server.zig#L1126-L1143) forever. Retrying cannot release resident capacity; only restarting the node can. That turns ordinary model switching into a service-liveness failure and contradicts the documented “temporary contention” semantics.
```

PR #412, opened 2026-07-30
Protect rebuild cursors against silent corruption

dovinmu:
Stacked on #371.

## What changed

- Store `rebuild.state` as a versioned envelope with a key length and CRC32.
- Sync the state file and its parent directory before considering an update durable.
- Reject malformed, truncated, oversized, or checksum-invalid cursors without clearing them.
- Route metadata status reporting through the same decoder instead of reading raw cursor bytes.
- Update metadata fixtures and add round-trip, mutation, and truncation regression tests.

## Why

The derived-index rebuild cursor was previously an unprotected raw key. A bit flip or partial write could turn a valid cursor into a later valid-looking key, permanently advancing the rebuild past documents that had not been indexed. The new format fails closed instead of silently publishing incomplete derived indexes.

## User impact

Corrupt rebuild progress now stops/quarantines the affected rebuild for repair rather than silently leaving search, vector, graph, or other derived indexes incomplete.

## Validation

- ReleaseSafe Antfly binary builds successfully.
- 6/6 focused rebuild-state tests pass in ReleaseSafe.
- 33/33 metadata-service consumer tests pass in ReleaseSafe.
- Formatting, AST checks, and `git diff --check` pass.
- A broader ReleaseSafe aggregate was run partially; many storage/indexing/recall suites passed, while independent split/merge simulation and broad unit failures remain on the stacked PR and were not addressed here.

### DIFF
--- a/zig/pkg/antfly/src/metadata/service.zig
+++ b/zig/pkg/antfly/src/metadata/service.zig
@@ -6166,6 +6166,14 @@ fn collectLocalStoreStatusReport(
         const millis = std.math.clamp(avg_progress * 1000.0, 0.0, 1000.0);
         report.backfill_progress_millis = @intFromFloat(millis);
     }
+    report.runtime_statuses = try collectQuarantinedBackfillRuntimeStatuses(
+        alloc,
+        backfill_markers,
+        store.store_id,
+        store.node_id,
+        true,
+    );
+    errdefer metadata_table_manager.freeRuntimeGroupStatusReports(alloc, report.runtime_statuses);
     report.group_statuses = try metadata_table_manager.cloneGroupStatuses(alloc, group_statuses);
     return report;
 }
@@ -6185,7 +6193,13 @@ fn collectSharedRootLocalStoreStatusReports(
 ) ![]metadata_table_manager.StoreStatusReport {
     _ = service;
     var reports = try alloc.alloc(metadata_table_manager.StoreStatusReport, stores.len);
-    errdefer alloc.free(reports);
+    errdefer {
+        for (reports) |report| {
+            metadata_table_manager.freeGroupStatuses(alloc, report.group_statuses);
+            metadata_table_manager.freeRuntimeGroupStatusReports(alloc, report.runtime_statuses);
+        }
+        alloc.free(reports);
+    }
     for (stores, 0..) |store, i| {
         reports[i] = .{
             .store_id = store.store_id,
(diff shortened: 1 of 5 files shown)


ajroetker:
[codex] Existing `rebuild.state` files are unversioned raw resume keys, but the new decoder requires the `AFRBST01` envelope and returns `InvalidRebuildState` for every legacy file. Upgrading while a rebuild is in progress would therefore quarantine/fail an otherwise valid rebuild. Please define a safe legacy-state policy (for example, discard the unverifiable cursor and restart the derived-index rebuild from scratch) and add an upgrade regression test covering a raw pre-v1 state file before merging.

ajroetker:
[codex] Review findings against base head `83cee3cb7`

I would not merge this as-is yet:

- **P1: corrupt state creates an unbounded metadata rescan loop.** `refreshStoreStatusBackfillMarkerResumeKeys` maps `error.InvalidRebuildState` to `resume_key = null`; `maybeRefreshStoreStatusBackfillMarkerCache` then sets `rescan_requested = true`. A persistent corrupt marker therefore causes a full root walk and decode attempt every lifecycle round. It is also reported as indistinguishable from a temporarily missing marker, so the promised quarantined/failure state is not observable. Use a tagged result (`absent`, `valid`, `corrupt`), cache `corrupt` without immediate rescans, and publish an explicit failed/quarantined derived-index status.
- **P2: status refresh now creates one I/O runtime per marker.** The old code created one `std.Io.Threaded` for the batch; `RebuildState.check()` creates/deinitializes its own runtime, and this PR calls it once per marker. Pass a shared `std.Io` through `check/update/clear` (with convenience wrappers only at process boundaries) so refresh remains O(markers) file reads without O(markers) runtime setup.
- **P1: the write path is not yet a fully atomic generation-safe replacement.** On rename `FileNotFound`, it recreates the parent and writes the destination directly, which can expose a partial file and can resurrect state after a concurrent index removal. The process-global non-atomic `temp_nonce` also does not provide collision safety across threads/processes. Create the temp file exclusively in the target directory, sync it, atomically replace, sync the parent, and treat a vanished parent as a fenced/stale owner rather than recreating it. The directory-sync unsupported-platform policy should also be explicit rather than returning an error after publication.

The versioned envelope, bounds, checksum, and file sync are the right integrity direction. Please rebase onto the current base and run CI after addressing the lifecycle/API shape; this draft currently has no reported checks.

ajroetker:
[codex] Addressed in `e44f5659e` and rebased onto the latest stacked base (`eb294fdfd`, including #413):

- Legacy raw cursors are now migrated to a versioned empty cursor, forcing a safe rebuild from scratch instead of quarantining a valid pre-v1 marker.
- Rebuild-state reads expose tagged `absent` / `valid` / `legacy` / `corrupt` results. Corrupt markers remain cached without an immediate lifecycle rescan loop and publish a per-index runtime status with `source: rebuild_state_quarantine` and `freshness: quarantined`.
- Metadata scanning and DB/index-manager call sites reuse shared `std.Io`; the allocating wrappers remain only as convenience boundaries.
- Writes now use cryptographically random, exclusively-created temp files in the target directory, file sync + atomic rename + parent sync, never recreate a vanished owner directory, and report stale-owner/durability-uncertain outcomes explicitly.

Regression coverage includes legacy upgrade/restart, corrupt marker caching and quarantine publication, vanished-owner fencing, corruption/truncation, full-text and sparse interrupted resumes, and the schema-progress marker probe. Focused results: rebuild-state 9/9, metadata service 35/35, table provisioner 1/1, full-text resume 1/1, sparse resume 1/1; Zig formatting and `git diff --check` pass.

PR #413, opened 2026-07-30
Fix MCP bridge leaks and gate A2A as experimental

dovinmu:
Stacked on #371.

## What changed

- Make converted httpx requests explicitly own and release their allocated header arrays across transaction and standalone bridges.
- Add `--experimental` to the data, metadata, and standalone runtimes while keeping MCP enabled by default.
- Keep A2A endpoints, agent cards, streaming, `/db/v1` aliases, and A2A discovery entries unavailable unless experimental features are enabled.
- Register standalone A2A routes only when opted in and route ARD through the shared server.

## Why

The standalone bridge allocated a header array for every converted request and never released it, so sustained MCP traffic could grow process memory. A2A is not ready to be a stable default surface, and its root and prefixed routes plus discovery need one consistent opt-in boundary.

## User impact

MCP remains available by default without the per-request bridge leak. A2A must be explicitly enabled with `--experimental`; when disabled, its endpoints and built-in discovery entries are hidden.

## Validation

- Focused API protocol and ARD tests: 5 passed, zero leaks.
- Standalone runtime tests: 42 passed, zero leaks.
- Focused data and metadata CLI parser tests passed, zero leaks.
- MCP and A2A library tests passed.
- `git diff --check` passed.


ajroetker:
[codex] Review findings against base head `83cee3cb7`

The converted-request ownership wrapper is a good fix and should be merged independently if needed, but I would not merge the A2A gate as-is:

- **P1: default MCP discovery still directs users to a disabled endpoint.** `protocol_adapters.zig` keeps a static `mcp_capabilities_json` and `describe_mcp_capabilities` description that tell clients to use the A2A query-builder. With this PRs default config, `/a2a` and the agent card return 404. The feature capability must flow into MCP capability generation so disabled surfaces are never advertised.
- **P1: standalone does not provide the enabled transport contract being tested.** `registerA2aRoutes` installs only root routes and sends requests through `protocolBridgeHandler -> ApiHttpServer.handle`, which uses the buffered `A2aSseSink`; it does not use `executeStreaming`/the live stream writer. Long `message/stream` calls can therefore buffer the whole event stream. It also does not register the `/db/v1/a2a` and `/db/v1/.well-known/...` aliases on the actual httpx router. The tests exercise `ApiHttpServer.handle` directly or a `RecordingServer`, so they do not catch either listener-level gap. Add a real-listener test and wire streaming plus aliases through the same feature-gated route registrar.
- **P2: `--experimental` is too broad and CLI-only for a durable product contract.** Enabling A2A should not implicitly enable every future experiment. Prefer one named capability (`a2a_enabled` / `--experimental-a2a`, also represented in config) consumed by route registration, direct dispatch, streaming, ARD discovery, and MCP discovery. That makes clustered/standalone behavior consistent and keeps future experiments independently controllable.

This draft also has no reported checks and is one base commit behind. Rebase it after the current A2A task-store hardening, then run focused real-listener tests plus normal CI.

PR #414, opened 2026-07-30
Protect rebuild cursors against silent corruption

ajroetker:
Replaces #412 now that its stacked base, #371, has merged.

## What changed

- Store `rebuild.state` as a versioned envelope with a key length and CRC32.
- Publish updates through an exclusive temporary file, sync the file, atomically rename it, and sync the parent directory.
- Reject malformed, truncated, oversized, or checksum-invalid cursors without clearing them.
- Migrate legacy raw cursors by safely restarting the rebuild from the beginning.
- Route metadata status reporting through the same decoder, cache corrupt markers without hot rescans, and publish a quarantined runtime status.
- Share one `std.Io` instance across rebuild state operations and add lifecycle regression coverage.

## Why

The derived-index rebuild cursor was previously an unprotected raw key. A bit flip or partial write could turn a valid cursor into a later valid-looking key, permanently advancing the rebuild past documents that had not been indexed. The new format fails closed instead of silently publishing incomplete derived indexes.

## User impact

Corrupt rebuild progress now stops and quarantines the affected rebuild for repair rather than silently leaving search, vector, graph, or other derived indexes incomplete.

## Validation

- 9/9 focused rebuild-state tests pass.
- 35/35 metadata-service tests pass.
- `zig fmt --check` passes for all changed Zig files.
- `git diff --check origin/main...HEAD` passes.

The branch was rebased onto `main` after #371 merged.

### DIFF
--- a/scripts/ci/zig-e2e-regression-loop.sh
+++ b/scripts/ci/zig-e2e-regression-loop.sh
@@ -59,6 +59,7 @@ fi
 
 if ((workers > 1)); then
   log_root="$(mktemp -d "${TMPDIR:-/tmp}/antfly-e2e-regression.XXXXXX")"
+  preserve_log_root=0
   pids=()
 
   # shellcheck disable=SC2329 # Invoked indirectly by the signal traps below.
@@ -70,8 +71,15 @@ if ((workers > 1)); then
     wait 2>/dev/null || true
     exit 130
   }
+  cleanup_worker_logs() {
+    if ((preserve_log_root == 0)); then
+      rm -rf -- "$log_root"
+    else
+      printf '\nPreserving E2E regression worker logs: %s\n' "$log_root" >&2
+    fi
+  }
   trap terminate_workers INT TERM
-  trap 'rm -rf "$log_root"' EXIT
+  trap cleanup_worker_logs EXIT
 
   for ((worker = 1; worker <= workers; worker++)); do
     env \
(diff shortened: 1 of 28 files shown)


dovinmu:
[codex]
```
• 1. Dense rebuild does not correctly resume from persisted progress. After an interrupted rebuild, Antfly should read its checkpoint and
     process only the four remaining artifacts. Instead, it processes eight—apparently repeating work that the checkpoint says was already
     completed. Besides wasting work, this suggests restart recovery may not honor its durable cursor and could distort progress or
     completion accounting.

  2. Chunk-backed dense rebuild performs or counts an extra rebuild. Documents split into chunks require careful accounting between the
     parent document and its derived chunk embeddings. The test expects two remaining artifacts to be rebuilt, but Antfly reports three.
     This suggests the new rebuild ownership logic may double-count a parent/chunk unit or fail to recognize one already-covered artifact,
     potentially delaying readiness or producing inaccurate coverage status.

  3. Graph reverse-index rebuild loses its restart checkpoint. The test deliberately interrupts reconstruction of the graph’s reverse-edge
     index, then expects a durable rebuild-state file to remain for restart. That file is missing, causing FileNotFound before recovery can
     be verified. This indicates graph rebuild state is still being written, looked up, or cleaned up under the wrong ownership/path, so an
     interrupted graph repair may restart incorrectly or be unable to resume.
```

PR #416, opened 2026-07-30
Tighten Zig and MemoryAF permission enforcement for 0.2.0

dovinmu:
## What changed

- require table `read` permission for Zig document scans
- add native Zig MCP regressions proving that read-only principals do not see write/admin tools and cannot use a visible read tool against an unauthorized table
- enforce private-memory visibility in the independently shipped MemoryAF package for explicit filters, direct reads, and graph-derived results

This PR contains no changes to the deprecated Go Antfly runtime. Hardening for that implementation belongs on the 0.1.x branch.

## Documentation

- document current Zig authentication coverage, internal-route trust boundaries, and the transaction authorization gap in `zig/AUTH.md`
- replace stale Go-parity MCP material with the native Zig MCP authorization and verification model in `zig/MCP.md`
- document the missing connection/egress/secret-use boundary for request-defined foreign PostgreSQL sources in `zig/CONNECTIONS.md`
- document MemoryAF's trusted identity integration, enforced visibility rules, and remaining entity-provenance limitation in `go/pkg/memoryaf/README.md`

## Why

The 0.2.0 sweep found a missing table-level check on the native Zig scan route and several MemoryAF paths that could surface another user's private memories.

The reported read-only MCP write bypass was in the deprecated Go adapter. Native Zig MCP already filters its advertised tools by effective permissions and routes built-in tool execution back through the normal authenticated HTTP handler, which enforces the exact target table, operation, and row filters. The added regressions make that distinction explicit.

Larger policy decisions—especially Zig transaction ownership/authorization, request-defined foreign DSNs, internal-route network boundaries, and private-derived MemoryAF entity metadata—are documented in their existing design or component files and remain intentionally outside this targeted-fix PR.

## Validation

- full MemoryAF tests passed with a temporary local shim for the pre-existing missing SDK `TransformOpTypeMin` alias; the shim is not included here
- `zig fmt --check zig/pkg/antfly/src/api/http_server.zig`
- `git diff --check`

The Zig suite was not built locally because that build is known to risk OOM on this machine; the added Zig regressions are left to CI.


ajroetker:
I found one authorization inconsistency that should be fixed before merge:

**Admin reads are still restricted to team + self-owned memories.** In `buildFilterQuery`, the explicit-private branch exempts admins, but the following default-visibility branch does not. As a result, an admin with no explicit visibility filter gets `(visibility=team OR created_by=<admin>)`, so `ListMemories`, `SearchMemories`, `GetStats`, and `ListSessions` omit private memories owned by other users. That contradicts the README statement that administrators retain the namespace-wide override, and it is inconsistent with the new `canReadMemory` behavior and the graph/direct-read paths. Please make the default admin predicate include all memory documents (while continuing to exclude entity documents where needed, especially from stats) and add admin regression coverage for these paths.

There is also an incomplete `$min` transform dependency in this PR. `memoryaf` now uses `client.TransformOpTypeMin`, but that constant is absent from the generated Go SDK, and the Zig batch parser currently explicitly rejects `$min`. Please implement `$min` end to end rather than relying on a local SDK shim:

- add it to the OpenAPI transform-op contract and regenerate/export it through the Go SDK;
- implement parsing, validation, serialization, and execution on the Zig server, including transaction/replication paths where transform operators are mapped;
- give it semantics symmetric with `$max` (numeric operands, update only when the operand is lower, and consistent missing/upsert behavior); and
- add client/server regression tests proving an SDK-issued `$min` transform succeeds and produces the expected stored value.

Without the temporary shim described in the PR body, `go test ./...` in `go/pkg/memoryaf` currently fails to compile with `undefined: client.TransformOpTypeMin`. With the shim, the MemoryAF suite passes.

PR #462, opened 2026-08-07
Fix enrichment retry lifecycle and repair UX

### LINKED ISSUE #460: Enrichment's retryable classifier is a closed allowlist with a terminal default: unenumerated transients permanently abandon documents while the index reports them pending
Enrichment decides whether a failure is transient by testing the error name against a closed allowlist, `isRetryableEnrichmentError` (`zig/pkg/antfly/src/storage/db/enrichment/enrichment_runtime.zig:456`). Sixteen names are enumerated; everything else falls through `else => false` and is treated as permanent.

This produces two opposite failure modes from one list, and in both the index status reports work in progress that will never progress.

Reproduced deterministically (2/2 identical runs) against the published `v0.2.0-rc.26` `Darwin_arm64` release asset. Script below is self-contained and downloads its own binary.

## Observed

A 32-document corpus. Twenty-four documents are embedded while the embedding endpoint returns HTTP 200 with a valid `Content-Length` and a JSON body cut mid-structure — a malformed response, not an unreachable one.

| | Unenumerated error | Enumerated error |
| --- | --- | --- |
| Injected | 200, honest `Content-Length`, truncated JSON body | HTTP 503 |
| Error name | `UnexpectedEndOfInput` | (enumerated transient) |
| `fatal_error_count` | **24** — one per document | **0** |
| `retryable_error_count` | 0 | — |
| `retrying` | `false` | `true` |
| Result | 24 documents permanently abandoned | **~54 embed requests/sec, unbounded** |

A truncated response body is transient by nature — the next request may well succeed. It permanently abandons every document it touches. Meanwhile an enumerated failure retries roughly 54 times per second indefinitely, with no backoff and no attempt budget.

## The index does not report either state

This is the part that matters most for consumers. After the 24 documents are permanently abandoned, the index reports:

```json
{
  "rebuilding": true,
  "backfill_state": "running",
  "backfill_active": true,
  "total_indexed": 8,
  "repair": null,
  "repair_degraded": false,
  "repair_issue_count": 0,
  "coverage": {
    "source_total": 32, "produced": 8, "skipped": 0,
    "terminal_failed": 24, "pending": 24,
    "complete": false, "healthy": false
  },
  "enrichment_runtime": {
    "fatal_error_count": 24, "retryable_error_count": 0,
    "retrying": false, "worker_failed": false
  }
}
```

It reports itself **rebuilding**, **actively backfilling**, with **24 documents pending** — permanently. There is no repair object, `repair_degraded` is `false`, and `repair_issue_count` is `0`. `coverage.terminal_failed: 24` and `coverage.pending: 24` are reported simultaneously for the same 24 documents.

A client polling index status cannot distinguish "still working" from "has permanently given up on three quarters of the corpus". The only available signal is `fatal_error_count > 0` alongside `retryable_error_count == 0` and `retrying: false` — which requires already knowing this failure mode exists in order to look for it.

## The state is permanent and survives restart

Restarting the process against the same data directory, with a **healthy** embedding endpoint:

```
after restart:  produced=8  terminal_failed=24  pending=24
                processed_requests=0  embed_batches_started=0
```

Zero requests processed. The worker starts, finds nothing to do, and the 24 documents stay abandoned with a working endpoint available. The state does not self-resolve on new writes.

Dropping and recreating the index fully recovers:

```
after drop+recreate:  produced=32  terminal_failed=0  total_indexed=32
                      complete=true  healthy=true  backfill_state="ready"
```

`fatal_error_count` remains at 24 across the recreate — it is a cumulative runtime counter and is not generation-scoped, so it does not reflect current health either.

### Why recreation is the only recovery

`recordIsolatedRequestError` (`:1760`) does not call `saveRuntimeStatusWithRetry`; the durable state is the coverage outcome rather than the runtime status blob. `markDerivedCoverageTerminalFailedForIndex` (`:8104`) → `markDerivedCoverageOutcomeForIndex` (`:8097`) → `applyCoverageOutcomeTransitions` writes marker and counter keys into the store. Each marker is namespaced by the index's coverage generation (`storage/db/catalog/index_manager.zig:7571`), and `applyCoverageOutcomeTransitions` discards any transition whose generation does not match the current one.

So the outcome is durable across restarts because it lives in the store, and is cleared by recreation because recreation changes the generation and orphans every prior marker. Restart-to-recover is structurally unavailable.

## Mechanism

`zig/pkg/antfly/src/storage/db/enrichment/enrichment_runtime.zig`

- `:456` — `isRetryableEnrichmentError`, the closed allowlist: 16 error names, then `else => false`.
- `:1760` — `recordIsolatedRequestError`, the per-request terminal path. Logs at `warn`, increments `fatal_error_count`, sets `retrying = false`, sets **`worker_failed = false`**, marks the index into `isolated_failed_indexes`, and stamps the document's coverage `terminal_failed`.
- `:1653` — `recordError`, a separate worker-level path that logs at `err` as `enrichment worker failed:` and sets `worker_failed = true`. This is **not** the path taken here.

Every call site of `recordIsolatedRequestError` (`:1987`, `:1992`, `:1997`, `:2002`, `:2198`, `:2217`, `:2224`, `:4542`, `:5164`) is correctly guarded by `if (isEnrichmentControlError(err) or isRetryableEnrichmentError(err)) return err;`. The guard is applied consistently — the issue is the allowlist's contents and its default, not a missing check.

**Diagnostic note.** Because the per-request path sets `worker_failed = false`, the worker keeps running and reports healthy, and no worker-failure line is ever logged. The evidence is N `warn`-level lines of the form `enrichment request failed index=… artifact=…: {errorName}`, one per abandoned document — 24 lines for 24 documents in this reproduction.

## Why these are one issue

The two behaviours come from one function and one list. Adding `UnexpectedEndOfInput` to the allowlist would resolve the reproduction above and leave both underlying problems intact: the next unenumerated transient would be equally fatal, and the unbounded-retry branch would be untouched.

There is direct precedent. #181 reported `ModelNotFound` classified as non-retryable; it was resolved by adding `ModelNotFound` to this allowlist, where it sits today. The same class of defect has recurred.

Suggested direction, offered rather than prescribed: invert the default. Treat unknown failures as retryable under a bounded budget — attempt count or elapsed time — and reserve terminal classification for errors positively known to be permanent, such as malformed configuration, unsupported model, or dimension mismatch. That addresses both branches at once. Independently of the classification decision, index status should distinguish work in progress from work permanently abandoned; `pending` and `terminal_failed` currently describe the same documents.

## Version

`isRetryableEnrichmentError` is byte-identical between `v0.2.0-rc.26` and `main` — same 16 names, same `else => false`. `UnexpectedEndOfInput` appears nowhere in `enrichment_runtime.zig` at either revision. This is live at HEAD, not a stale-release artifact.

Possibly related: #442 (latched rebuild flag), #384 (enrichment items pending forever with no error and no retry-exhaustion — plausibly this same behaviour observed from the outside), #181 (above).

## Reproduction

macOS arm64; change `ASSET_PATTERN` for other hosts. Requires `gh`, `curl`, `jq`, `python3`, `tar`. Exits 0 when all three arms reproduce, and aborts loudly if the control arm fails so that a harness fault cannot be mistaken for a defect.

<details>
<summary><code>repro-enrichment-terminal-classification.sh</code></summary>

```bash
#!/usr/bin/env bash
set -euo pipefail

ANTFLY_VERSION="${ANTFLY_VERSION:-v0.2.0-rc.26}"
ASSET_PATTERN="${ASSET_PATTERN:-*Darwin_arm64.tar.gz}"
DOC_COUNT="${DOC_COUNT:-24}"
STORM_WINDOW_S="${STORM_WINDOW_S:-60}"
OBSERVE_S="${OBSERVE_S:-90}"

WORK="$(mktemp -d "${TMPDIR:-/tmp}/antfly-enrich-repro.XXXXXX")"
ANTFLY_PID=""; EMBED_PID=""
cleanup() {
  [ -n "$ANTFLY_PID" ] && kill "$ANTFLY_PID" 2>/dev/null || true
  [ -n "$EMBED_PID" ] && kill "$EMBED_PID" 2>/dev/null || true
  wait 2>/dev/null || true
  echo; echo "artifacts retained in: $WORK"
}
trap cleanup EXIT
say() { printf '\n=== %s ===\n' "$*"; }
die() { printf 'FAIL: %s\n' "$*" >&2; exit 1; }
free_port() { python3 -c 'import socket;s=socket.socket();s.bind(("127.0.0.1",0));print(s.getsockname()[1]);s.close()'; }

for tool in gh curl jq python3 tar; do
  command -v "$tool" >/dev/null 2>&1 || die "missing required tool: $tool"
done

say "fetching antfly $ANTFLY_VERSION release asset"
gh release download "$ANTFLY_VERSION" --repo antflydb/antfly \
  --pattern "$ASSET_PATTERN" --dir "$WORK" || die "could not download asset"
tar -xzf "$WORK"/*.tar.gz -C "$WORK"
ANTFLY_BIN="$(find "$WORK" -type f -name antfly -perm -u+x | head -1)"
[ -n "$ANTFLY_BIN" ] || die "no antfly binary inside the release asset"

# Mode-switchable embedding endpoint speaking the OpenAI embeddings wire protocol.
# Behaviour is switched at runtime by writing a word into $WORK/mode. The index must
# be created while the endpoint is healthy: index creation validates the embedder and
# refuses to create an index against a dead endpoint, so the fault must arrive after.
#
#   healthy   200, well-formed response
#   truncate  200, application/json, HONEST Content-Length, body cut mid-JSON.
#             The read completes; the parse does not. Lying about Content-Length
#             instead would surface as a read error, and UnexpectedReadFailure IS
#             enumerated as retryable -- that would exercise the wrong arm.
#   refuse    503 with a JSON error body -- an enumerated transient
EMBED_PORT="$(free_port)"; EMBED_URL="http://127.0.0.1:${EMBED_PORT}"
echo healthy > "$WORK/mode"; : > "$WORK/embed-requests.log"

cat > "$WORK/embedder.py" <<'PYEOF'
import json, os, sys, time
from http.server import BaseHTTPRequestHandler, ThreadingHTTPServer
WORK, PORT, DIMS = sys.argv[1], int(sys.argv[2]), int(sys.argv[3])
def mode():
    try:
        with open(os.path.join(WORK, "mode")) as fh: return fh.read().strip()
    except OSError: return "healthy"
def vector_for(text):
    h = abs(hash(text))
    return [((h >> (8 * i)) & 0xFF) / 255.0 for i in range(DIMS)]
class Handler(BaseHTTPRequestHandler):
    def do_POST(self):
        if self.path != "/v1/embeddings": self.send_error(404); return
        raw = self.rfile.read(int(self.headers.get("Content-Length", "0")))
        payload = json.loads(raw.decode("utf-8") or "{}")
        inputs = payload.get("input", [])
        if isinstance(inputs, str): inputs = [inputs]
        m = mode()
        with open(os.path.join(WORK, "embed-requests.log"), "a") as fh:
            fh.write("%.6f %s %d\n" % (time.time(), m, len(inputs)))
        if m == "refuse":
            body = json.dumps({"error": {"message": "capacity exhausted",
                                         "type": "server_error"}}).encode()
            self.send_response(503)
            self.send_header("Content-Type", "application/json")
            self.send_header("Content-Length", str(len(body)))
            self.end_headers(); self.wfile.write(body); return
        full = json.dumps({
            "object": "list", "model": payload.get("model", "text-embedding-3-small"),
            "data": [{"object": "embedding", "index": i, "embedding": vector_for(str(t))}
                     for i, t in enumerate(inputs)],
            "usage": {"prompt_tokens": max(1, len(inputs)),
                      "total_tokens": max(1, len(inputs))}}).encode()
        if m == "truncate":
            full = full[: max(24, len(full) // 2)]
        self.send_response(200)
        self.send_header("Content-Type", "application/json")
        self.send_header("Content-Length", str(len(full)))
        self.end_headers(); self.wfile.write(full)
    def log_message(self, *a): pass
ThreadingHTTPServer(("127.0.0.1", PORT), Handler).serve_forever()
PYEOF

python3 "$WORK/embedder.py" "$WORK" "$EMBED_PORT" 3 &
EMBED_PID=$!
for _ in $(seq 1 50); do
  curl -fsS -m 2 -X POST "$EMBED_URL/v1/embeddings" -H 'Content-Type: application/json' \
    -d '{"model":"text-embedding-3-small","input":["probe"]}' >/dev/null 2>&1 && break
  sleep 0.2
done

API_PORT="$(free_port)"; HEALTH_PORT="$(free_port)"; API="http://127.0.0.1:${API_PORT}"
start_antfly() {
  "$ANTFLY_BIN" standalone --host 127.0.0.1 --port "$API_PORT" \
    --health-port "$HEALTH_PORT" --data-dir "$WORK/data" > "$WORK/antfly.log" 2>&1 &
  ANTFLY_PID=$!
  for _ in $(seq 1 120); do
    curl -fsS -m 2 "$API/db/v1/status" >/dev/null 2>&1 && return 0
    sleep 0.5
  done
  tail -40 "$WORK/antfly.log" >&2; die "antfly did not become ready"
}
say "starting antfly standalone"; start_antfly

TABLE="enrich_classification"; INDEX="document_vectors"
raw_indexes() { curl -fsS -m 10 "$API/db/v1/tables/$TABLE/indexes" 2>/dev/null; }
# Status is an array of IndexStatus{config,status,shard_status}; counters live on
# .status. Select on .config.name -- a recursive search matches config first.
index_status() {
  raw_indexes | jq -c --arg n "$INDEX" '
    (if type == "array" then .[] else (.indexes // [])[] end)
    | select(.config.name == $n) | .status'
}
create_index() {
  curl -fsS -m 30 -X POST "$API/db/v1/tables/$TABLE/indexes/$INDEX" \
    -H 'Content-Type: application/json' \
    -d "$(jq -n --arg n "$INDEX" --arg url "$EMBED_URL" '{
          name: $n, type: "embeddings", field: "content", dimension: 3,
          embedder: { provider: "openai", model: "text-embedding-3-small", url: $url }}')"
}
insert_docs() {
  local prefix="$1" count="$2" inserts
  inserts="$(python3 - "$prefix" "$count" <<'PY'
import json, sys
prefix, count = sys.argv[1], int(sys.argv[2])
print(json.dumps({"inserts": {
    f"{prefix}-{i:04d}": {"content": f"document {prefix} number {i} with enough text to embed"}
    for i in range(count)}}))
PY
)"
  curl -fsS -m 60 -X POST "$API/db/v1/tables/$TABLE/batch" \
    -H 'Content-Type: application/json' -d "$inserts" >/dev/null
}

say "creating table and index (endpoint healthy -- create-time probe must pass)"
curl -fsS -m 30 -X POST "$API/db/v1/tables/$TABLE" \
  -H 'Content-Type: application/json' -d '{"num_shards":1}' >/dev/null
create_index || die "index creation failed while the endpoint was healthy"

say "control: healthy backfill produces vectors"
insert_docs healthy 8
produced=0
for _ in $(seq 1 90); do
  produced="$(index_status | jq -r '
    [ (.coverage.produced_count // .coverage.produced // 0),
      (.total_indexed // 0), (.doc_count // 0) ] | max // 0')"
  [ -n "$produced" ] && [ "$produced" -gt 0 ] 2>/dev/null && break
  sleep 1
done
if ! [ "${produced:-0}" -gt 0 ] 2>/dev/null; then
  raw_indexes | jq . >&2; cat "$WORK/embed-requests.log" >&2
  die "control arm never produced vectors -- harness fault, not a defect"
fi
echo "control produced: $produced"

say "ARM A: unenumerated transient (truncated JSON body)"
echo truncate > "$WORK/mode"
insert_docs poisoned "$DOC_COUNT"
fatal=0; retryable=0; retrying="?"
for _ in $(seq 1 "$OBSERVE_S"); do
  s="$(index_status)"
  fatal="$(echo "$s" | jq -r '.enrichment_runtime.fatal_error_count // 0')"
  retryable="$(echo "$s" | jq -r '.enrichment_runtime.retryable_error_count // 0')"
  retrying="$(echo "$s" | jq -r '.enrichment_runtime.retrying // false')"
  [ "$fatal" -gt 0 ] 2>/dev/null && break
  sleep 1
done
echo "fatal_error_count=$fatal retryable_error_count=$retryable retrying=$retrying"
index_status | jq . > "$WORK/arm-a-status.json"
echo "log lines 'enrichment request failed': $(grep -c 'enrichment request failed' "$WORK/antfly.log" || true)"
grep -oE 'enrichment request failed[^"]*' "$WORK/antfly.log" 2>/dev/null | tail -3 || true
[ "$fatal" -gt 0 ] 2>/dev/null || die "ARM A did not reproduce"
echo "ARM A REPRODUCED"

say "ARM B: does that state survive a restart?"
kill "$ANTFLY_PID" 2>/dev/null || true; wait "$ANTFLY_PID" 2>/dev/null || true
ANTFLY_PID=""; echo healthy > "$WORK/mode"
start_antfly; sleep 10
index_status | jq . > "$WORK/arm-b-after-restart.json"
jq -c '{fatal: .enrichment_runtime.fatal_error_count,
        processed: .enrichment_runtime.processed_requests,
        produced: .coverage.produced, terminal_failed: .coverage.terminal_failed,
        pending: .coverage.pending}' "$WORK/arm-b-after-restart.json"

say "ARM B: does dropping and recreating the index clear it?"
curl -fsS -m 30 -X DELETE "$API/db/v1/tables/$TABLE/indexes/$INDEX" >/dev/null || true
sleep 3; create_index || die "index re-creation failed"; sleep 15
index_status | jq . > "$WORK/arm-b-after-recreate.json"
jq -c '{produced: .coverage.produced, terminal_failed: .coverage.terminal_failed,
        total_indexed, backfill_state, complete: .coverage.complete,
        healthy: .coverage.healthy}' "$WORK/arm-b-after-recreate.json"

say "ARM C: enumerated transient (HTTP 503) -- retry rate over ${STORM_WINDOW_S}s"
echo refuse > "$WORK/mode"
insert_docs storm "$DOC_COUNT"
before="$(grep -c ' refuse ' "$WORK/embed-requests.log" 2>/dev/null || echo 0)"
sleep "$STORM_WINDOW_S"
after="$(grep -c ' refuse ' "$WORK/embed-requests.log" 2>/dev/null || echo 0)"
delta=$(( after - before ))
echo "refused embed requests in ${STORM_WINDOW_S}s: $delta (~$(python3 -c "print(f'{$delta/$STORM_WINDOW_S:.1f}')")/sec)"
index_status | jq -c '{fatal: .enrichment_runtime.fatal_error_count,
                       retrying: .enrichment_runtime.retrying, backfill_state}'
```

</details>

### Expected output

```
control produced: 8

=== ARM A: unenumerated transient (truncated JSON body) ===
fatal_error_count=24 retryable_error_count=0 retrying=false
log lines 'enrichment request failed': 24
enrichment request failed index=document_vectors artifact=document_vectors: UnexpectedEndOfInput

=== ARM B: does that state survive a restart? ===
{"fatal":24,"processed":0,"produced":8,"terminal_failed":24,"pending":24}

=== ARM B: does dropping and recreating the index clear it? ===
{"produced":32,"terminal_failed":0,"total_indexed":32,"backfill_state":"ready",
 "complete":true,"healthy":true}

=== ARM C: enumerated transient (HTTP 503) -- retry rate over 60s ===
refused embed requests in 60s: 3216 (~53.6/sec)
{"fatal":24,"retrying":true,"backfill_state":"retrying"}
```

Two runs on the same host produced identical counts and the same error name; the measured retry rate was 53.6/sec and 54.3/sec. Retry rate is machine- and batch-dependent, but is unbounded in both.

## Environment

macOS 15 arm64, `antfly v0.2.0-rc.26` published `Darwin_arm64` release asset, `standalone` mode with local metadata (raft disabled), three-dimensional embeddings index against an external OpenAI-protocol endpoint. No models are downloaded and no API key is required.

ajroetker:
Fixes #460

## Summary

- apply one bounded retry lifecycle to worker, foreground, and startup catch-up execution
- persist request-scoped retry identity so one poisoned document cannot consume the next documents retry budget
- keep inline provider retries separate from durable worker attempts and retry pipeline failures without charging document budgets
- treat max attempts as a true total-attempt bound and persist exact source sequence and generation failure telemetry
- park exhausted failures for every transitive vector, full-text, and graph consumer; producer-only failures remain globally repairable without pretending an enrichment name is an index
- keep generation attempts and errors separate from later repair attempts and errors
- support forced regeneration for copy, generator, reader, transcriber, extractor, and document-extraction asset producers
- define one policy-aware coverage assessment shared by status and repair reporting
- report committed but incomplete work as indexes_rebuilt plus indexes_degraded and debt_remaining; reserve repaired for complete, debt-free outcomes
- apply the same outcome classification to normal completion, crash reconciliation, and idempotent repair-job replay
- return terminal foreground failures as worker failures rather than a misleading retry-in-progress response
- regenerate Zig, Go, TypeScript, and Python SDK bindings

## Operator UX

A completed rebuild no longer implies a successful repair:

- indexes_rebuilt means replacement work committed
- indexes_degraded identifies selected indexes observed in a degraded state, including a committed generation that remains incomplete
- debt_remaining is the authoritative signal that known or indeterminate work remains
- repaired is emitted only when the generation satisfies its configured coverage policy and no repair debt remains

Strict, partial, and best-effort coverage now use the same completion rules in status and repair paths. Missing or corrupt counter tuples fail closed as indeterminate debt; external and direct-field vector indexes do not incorrectly require derived-enrichment counters.

Repair issues expose generation_attempts and generation_error without overloading the repair-attempt fields attempts and last_error.

## Tests

- make generate
- zig build db-enrichment-test (107 passed)
- zig build api-table-writes-docid-test (72 passed)
- zig build lib-api-derived-coverage-test (36 passed)
- zig build lib-api-artifact-reprocess-jobs-test (24 passed)
- zig build openapi-root-check
- Go SDK tests
- TypeScript SDK typecheck
- Python SDK tests (15 passed)
- Zig formatting and git diff checks

The aggregate generated check reaches and passes OpenAPI generation, but its Snowball check cannot run because zig/deps/snowball is absent from this worktree.


drewlanenga:
The classification inversion and the `settled`/`uncovered`/`pending` split both look right to me, and parking exhausted failures into the artifact repair ledger directly addresses the second comment on #460 — `repair/issues` will now surface these documents instead of reporting zero debt.

One adjacent surface appears untouched, and I think it is worth deciding on deliberately rather than by omission: **`repair/run` reports success over a generation that is known to be incomplete.**

### Baseline measurement

Measured on `v0.2.0-rc.26` (pre-PR), using the reproduction from #460. After 24 of 32 documents were terminally abandoned, with the embedding endpoint returned to healthy and *before* any drop-and-recreate:

```
POST /repair/issues -> {"issues": 0}

POST /repair/run {"target":"index","index":"document_vectors","force":true,"limit":1}
  -> {"scanned":1, "groups_scanned":1, "reprocessed":0, "repaired":0,
      "missing_source_docs":0, "failed":0, "unsupported":0,
      "unresolved":1, "in_progress":1, "indexes_rebuilt":0,
      "indexes_degraded":0, "controls_applied":0, "limit":1,
      "has_more":false, "debt_remaining":true}

server log:
  index repair complete ... duration_ms=91 attempted=true repaired=true
                             pending=false busy=false

index afterwards:
  {"total_indexed":8, "backfill_state":"running", "repair":null,
   "repair_degraded":false, "repair_issue_count":0,
   "coverage":{"source_total":32, "produced":8, "terminal_failed":24,
               "pending":24, "complete":false}}
```

The repair genuinely ran — `attempted=true`, 91ms — and reported `repaired=true` with `indexes_degraded: 0`, over an index holding 8 of 32 documents.

### What this PR changes, and what it does not

It fixes the two inputs that made the above worst:

- `repair/issues` will no longer report zero, because `recordEnrichmentRequestFailure` parks these into the ledger.
- Index status will no longer read `pending: 24` alongside `terminal_failed: 24`, and `degraded` is now computed as `observation_complete and replay_current and counters_valid and all_sources_terminal and terminal_failed > 0`.

What it does not appear to change: the repair run path itself. `indexes_degraded`, `indexes_rebuilt`, `repaired`, and the `attempted=true repaired=true` log line are untouched in this diff, and `api/repair_jobs.zig` is not among the changed files.

So after this PR, an index in the degraded state that the PR newly defines should still be able to answer a forced repair with `repaired=true` and `indexes_degraded: 0`, while `coverage.complete` is false and the new `degraded` flag is true.

### The question

Now that `degraded` has a precise definition in index status, should the repair run result agree with it? Concretely:

- Should `indexes_degraded` increment when a repair completes over a generation whose coverage is settled-but-incomplete?
- Should `repaired=true` be reserved for a repair that produced a complete generation, with a distinct disposition for "rebuilt what was available, debt remains"?

`debt_remaining: true` was already correct in the measurement above, so the information is present in the response — it is the summary verbs (`repaired`, `indexes_degraded`) that disagree with it. An operator checking whether a repair worked is most likely to read those.

Not blocking, and possibly out of scope for this PR by design — but if it is deliberately deferred it seems worth saying so, since #460's central complaint was that every layer reported healthy while the corpus was incomplete, and this is the one layer where that is still reachable.

I have not re-run the reproduction against this branch; the numbers above are the rc.26 baseline, and the claims about this PR are read from the diff rather than measured. Happy to re-run against the branch if that would help.

ajroetker:
Addressed in 726e19b9e.

The long-term UX contract is now:

- `indexes_rebuilt`: replacement work committed
- `indexes_degraded`: the committed generation has settled-but-incomplete coverage
- `debt_remaining`: known or indeterminate repair debt remains
- `repaired`: complete generation with no remaining debt

The implementation uses one policy-independent coverage-health evaluator for status and repair reporting. The repair completion path checks generation health, exact coverage counters, and the durable artifact-repair summary before setting `repaired`. The same classifier also runs on crash-idempotent redispatch, so a durable completion marker cannot be mistaken for proof of healthy coverage.

The regression reproduces shared enrichment exhaustion, clears the new ledger entries to model legacy debt, performs the forced rebuild, and verifies both the initial response and replay return `indexes_rebuilt=1`, `indexes_degraded=1`, `repaired=0`, and `debt_remaining=true`.

Also included from the broader review: unified startup/foreground retry accounting, exact per-consumer debt for shared producers, separate generation versus repair telemetry, and forced regeneration for every asset producer type. Focused Zig suites pass 106 + 72 + 36 + 24 tests; generated Go, TypeScript, and Python SDK checks pass as well.

ajroetker:
Follow-up hardening is in 7f9053140.

This closes the remaining review findings:

- durable retry attempts are keyed to a stable request/batch fingerprint, so exhausted work cannot consume the next documents budget after continuation or restart
- inline provider retries no longer consume durable worker attempts, while unattributed pipeline failures remain retryable without being converted into document terminal coverage
- asset/chunk failure fanout now resolves transitive dense, sparse, full-text, and graph consumers; producer-only debt uses the global repair scope instead of exposing an enrichment name as index_name
- repair completion is policy-aware for strict, partial, and best_effort coverage, treats missing counter tuples as indeterminate debt for non-empty sources, and exempts external/direct-field indexes that do not own derived counters
- normal completion, activated-generation crash reconciliation, and idempotent redispatch all use the same finalizer
- foreground terminal failures now return EnrichmentWorkerFailed rather than EnrichmentRetryInProgress

Regression coverage includes request identity/reset behavior, v3 durable retry-state migration, transitive vector/full-text consumer resolution, graph/default-full-text consumers, and shared policy assessment. Current focused results: enrichment 108/108, table writes 72/72, derived coverage 36/36, repair jobs 24/24; make generate and the OpenAPI root check pass.

ajroetker:
Follow-up 53563fba4 removes the legacy runtime-status reader and migration regression, per the major-release compatibility policy. The retry status now has one canonical 62-byte format and rejects older layouts. The enrichment suite remains green at 107/107.

PR #469, opened 2026-08-10
Port bounded Zig runtime graph and parallel memory scheduling

ajroetker:
## Summary

- ports the high-value PR #450 ReleaseFast partition into four static runtime units: API, distributed/application storage, inference, and remote CLI
- links the single PIC distributed storage archive into the executable and both C APIs, with section GC and production LSM-only reachability
- preserves narrow API, standalone inference, and restore-staging bridges while removing broad cross-runtime imports
- shards the monolithic storage unit-test codegen graph and gives storage, metadata, and remaining aggregate compile/run steps explicit RSS claims
- removes CI reliance on -j1; the Zig 0.16 max-RSS scheduler now uses 80% of the cgroup limit with the wake-accounting backport

## Verification

- clean ARM64 Linux musl ReleaseFast install + lite-capi build, linked runtimes, production LSM-only, normal concurrency, 20 GiB scheduler budget
- unit-storage-test with HA disabled, normal concurrency, patched runner, 12 GiB scheduler budget
- linked Debug install build
- runtime, codegen, and API-kernel graph boundary checks
- 15 graph analyzer and max-RSS patch unit tests
- generated OpenAPI, Snowball, and C ABI checks

## PR #467 job note

The referenced job reached about 17.4 GB cgroup peak but recorded no OOM event; its terminal failure was the document-scan authorization assertion in api.http_server. This PR addresses the build-memory exposure that job demonstrated without masking or changing that separate functional failure.


dovinmu:
```
 1. Release packaging fails for explicit Darwin targets. The new production link-root modules propagate
    Foundation, Metal, Accelerate, and MetalPerformanceShaders, but do not attach the macOS SDK
    framework/library search paths. Consequently, the aarch64-macos ReleaseFast packaging build fails with
    searched paths: none. The affected roots are libantfly_link_mod, antfly_main_mod, each RuntimeLibraryUnit,
    and each RuntimeArtifactRole. Calling the existing addMacosSdkPaths helper for those four categories fixes
    the build.
 2. Graph transform errors lose their identity across the runtime boundary. A $push to an existing graph edge
    array should be rejected as HTTP 400, as on the pre-PR base. PR #469 instead returns HTTP 500 batch failed.
    The underlying UnsupportedTransformOperation is absent from the stable runtime error ABI, falls through to
    .internal/.none, and is reconstructed as RuntimeBoundaryFailure. The existing edge remains intact, so this
    is an error-translation regression rather than edge corruption.
```

ajroetker:
```
 ## Findings

  - P1 — The runtime boundary is still Zig-layout-dependent. .worktrees/port-pr450-build-memory/zig/pkg/antfly/src/runtime_callback_abi.zig:40 transports std.meta.ArgsTuple, arbitrary Zig payloads, allocators, slices, domain structs, and Zig
    function pointers through anyopaque. Only the error status is language-neutral. This does not satisfy the intended stable ABI and prevents safe independent reuse/versioning of runtime archives. Each operation needs explicit extern struct
    inputs/outputs or serialized bytes and C-calling-convention entry points.

  - P1 — Replicated restore cleanup bypasses the new error ABI. .worktrees/port-pr450-build-memory/zig/pkg/antfly/src/api/restore_jobs.zig:446 invokes persistence.vtable.delete_many directly. In production this returns a compilation-local
    anyerror from distributed into the API unit. Use persistence.deleteMany(...) and add a production-boundary test where expired-row deletion fails.

  - P1 — Release packaging no longer has a real memory bound. .worktrees/port-pr450-build-memory/scripts/packaging/build_zig_release_archive.sh:171 defaults to a fixed ~19.5 GiB scheduler budget, while neither release workflow overrides it. That
    permits the 11 GiB distributed, 6 GiB API, and 2 GiB CLI claims to run together, potentially exceeding a 15 GiB runner after removing -j1. Reuse the repository’s cgroup/host-aware budget detection instead.

  - P2 — Storage sharding can silently omit tests. .worktrees/port-pr450-build-memory/zig/build.zig:6810 replaces the comprehensive storage. selection with a manually maintained prefix allowlist. Adding a new storage module does not fail CI; its
    tests simply never enter a shard. Use explicit shard roots that collectively import the storage test surface, or add a coverage audit comparing discovered tests against the shard union.

  The current PR head is fully green, and all 24 new build-helper tests pass locally. The tag-only release jobs and replicated restore failure path are not covered by those checks. I would not merge until the three P1 findings are addressed.
```

PR #472, opened 2026-08-11
fix(zig): resume managed embedding backfill after restart

ajroetker:
## Summary

- restore metadata-owned managed embedding providers during startup catch-up
- keep a short-lived startup probe from terminalizing transient provider failures at the persisted retry-budget boundary
- add a focused unit regression and include it in the API table-write test gate

## Testing

- zig build -Dedition=full install -fincremental
- focused startup catch-up unit regression: 1 passed
- serial repository regression loop: 10/10 passed
- concurrent repository regression loop: 3 workers x 5 iterations, 15/15 passed
- git diff --check

## Context

Follow-up for #347. Reproduced from https://github.com/antflydb/antfly/actions/runs/31531357544/job/93912435099?pr=347.


ajroetker:
Superseded by commit 96917f14e on #469, where the fix can run against the combined branch and its main-targeting CI.

PR #476, opened 2026-08-12
Fix inference CI routing, model listing, and memory admission

timkaye11:
[codex] 

## Summary

  - Fix Linux live-memory admission under shared-host pressure and add bounded sparse-embedding batch
  fallback.
  - Publish stable `owner/model` identifiers instead of internal cache paths or variant hashes.
  - Make `/ai/v1/extract` the canonical classification API:
    - Dispatch GLiNER and zero-shot NLI classifiers.
    - Add per-taxonomy `hypothesis_template` and `top_k`.
    - Preserve input IDs and return usage.
    - List classifiers under `models.extractors`.
    - Keep SDK `classify()` as an `/extract` wrapper.
    - Do not expose a public `/classify` operation.
  - Correct speculative-decoding tests to require a distinct draft model without adding a dedicated CI
  job.
  - Stabilize short-generation and vector-cache tests.
  - Split inference and Antfly full E2E into independent legs while retaining the required `e2e-full`
  aggregate.

  ## Validation

  - Focused Zig tests: 5/5 passed.
  - Non-Metal inference build passed.
  - Go SDK tests passed.
  - TypeScript: 47 tests and typecheck passed.
  - Python SDK: 12 tests passed.
  - Model-helper tests: 42 passed.
  - OpenAPI/code generation and workflow validation passed.


ajroetker:
[codex]

Reviewed the updated classification UX against the canonical `/ai/v1/extract` design. The core design is incorporated: there is no public `/classify` operation, NLI and GLiNER dispatch through `/extract`, `hypothesis_template` and `top_k` are taxonomy-local, input IDs and usage are preserved, SDK `classify()` remains an `/extract` wrapper, and classifier models are presented under `models.extractors`.

Two findings to address before this is complete:

1. Fixed-label classifiers are advertised as extraction-capable but cannot be invoked. `modelSupportsCapability` infers `classification` for every internal classifier, so all are listed under `models.extractors`; the request schema requires caller labels and the runtime rejects fixed-label classifiers. For this PR, only GLiNER and classifiers proven `zero_shot` should be exposed as extraction classifiers. Fixed-label support can follow with optional labels and manifest `id2label` semantics.

2. `options.threshold` is currently applied to single-label results before `top_k`. This contradicts the documented contract that single-label returns its highest-ranked result(s), while threshold selects multi-label results. A high threshold can produce an empty single-label classification. Apply threshold only when `multi_label` is true and add a regression test where every single-label score is below threshold.

Non-blocking cleanup: NLI usage undercounts the text-by-label/hypothesis work, and the old unregistered `classifyText` handler plus `Classify*` component schemas remain as dead surface.

I am following up with the two behavioral fixes and focused tests on this PR branch.

ajroetker:
[codex]

Follow-up fixes pushed in `8a27ddce7`:

- Only zero-shot-capable internal classifier models are now exposed under `models.extractors`; fixed-label classifiers are no longer advertised with an unsupported request contract.
- `options.threshold` now filters only multi-label classifications. Single-label taxonomies always return their ranked `top_k` results.
- Focused regression coverage was updated for both behaviors.

Validation: `zig build inference-test -- --test-filter "taskMatchesModelListing" --test-filter "classification extraction applies"` — 3 selected, 3 passed.

ajroetker:
[codex] Re-review findings at `8a27ddc`:

- **P1 — merge conflict:** `main` has advanced by 10 commits and GitHub reports this PR as `mergeable: false` / `dirty`. The conflict is in the generated Go OpenAPI client, so the branch needs to be updated from `main` and generated clients refreshed.
- **P2 — NLI usage accounting:** `/extract` currently counts each input text and candidate label once, but the NLI pipeline tokenizes every text-label hypothesis pair. `usage.prompt_tokens` therefore undercounts work by roughly the text × label product and omits the rendered hypothesis template.
- **Validation gap:** the current head has no GitHub check runs or commit statuses.

The classifier UX otherwise matches the intended design: classification is canonicalized under `/extract`, callable classifiers are listed under `models.extractors`, fixed-label classifiers are hidden, per-taxonomy `hypothesis_template` / `top_k` are supported, IDs are preserved, and single-label vs. multi-label filtering follows the documented contract.

ajroetker:
[codex] Fixed in `82bef16ad` and pushed to the PR branch.

- Merged current `main` and regenerated `openapi.yaml` plus the Go SDK client, resolving the merge conflict.
- Changed the NLI classification pipeline to report exact non-padding tokens from every encoded text-label hypothesis pair; `/extract` now aggregates those authoritative counts across taxonomies.
- Added coverage asserting the Cartesian-pair token count.

Validation:
- `zig build inference-test`: 3,080 passed, 17 skipped.
- Focused classifier tests: 5/5 passed.
- `GOWORK=off go test ./...` in `go/pkg/sdk`: passed.
- GitHub now reports the PR mergeable; fresh checks have started and are pending.

PR #479, opened 2026-08-12
ci: evict obsolete PR workflow runs

ajroetker:
## Summary

- cancel active `pull_request` workflow runs from superseded heads whenever a PR is synchronized
- cancel active `pull_request` workflow runs associated with a PR when it closes, including the current head
- serialize synchronize, close, and reopen cleanup per PR so a reopen supersedes an in-flight close cleanup
- give ordinary cancellation a bounded 15-second grace period, then force-cancel only runs whose active status stayed stable
- reject closed or stale PRs before scarce ARC capacity is spent on Zig base and E2E setup; scheduled Zig coverage is unchanged

## Safety

The `pull_request_target` workflow never checks out or executes PR code. Its token is limited to `actions: write` and `contents: read`. Synchronize cleanup is restricted to stale heads for the same PR/ref, close cleanup is restricted to runs associated with that PR or its exact closing head, and live PR state/head is checked before cancellation and again immediately before force escalation. Only `pull_request` runs are queried, so post-merge `push` and `merge_group` runs are outside its scope.

The heavy Zig admission guards require both an open PR and the exact event head before checkout. The Zig base guard runs only for `pull_request` events, leaving the weekly scheduled job intact.

## Validation

- `actionlint .github/workflows/cancel-superseded-pr-runs.yml .github/workflows/zig-tests.yml`
- `git diff --check`


bpopadiuk (dismissed):
Codex (GPT-5): One race in the superseded-run canceller needs fixing before merge.

.github/workflows/cancel-superseded-pr-runs.yml
```diff
@@ -0,0 +1,90 @@
...
+              {
+                owner,
+                repo,
+                event: "pull_request",
+                branch: pull.head.ref,
+                per_page: 100,
+              },
+            );
+            const staleRuns = runs.filter((run) =>
+              run.head_sha !== pull.head.sha &&
```
bpopadiuk: This compares active runs against the `synchronize` event payload SHA, which may be stale by the time this queued cancellation job starts. If push A queues this job and push B lands before it runs, the A job can see B's active run as `run.head_sha !== pull.head.sha` and cancel the current CI. Consider fetching the live PR head first, or exiting when the event SHA is no longer current.

bpopadiuk (commented):
Codex (GPT-5): One informational note on the superseded-run canceller race window.

bpopadiuk (approved):

PR #482, opened 2026-08-13
Decode compressed remote content before extraction

dovinmu:
> ## Summary
> - decode gzip and deflate HTTP response bodies before returning downloaded content
> - treat standards-compliant zlib-wrapped `deflate` as the primary form while retaining raw-deflate compatibility
> - decode percent-encoded `data:` URLs and enforce download limits on decoded bytes
> - cover gzip, zlib-wrapped deflate, and raw deflate through buffered and writer HTTP paths
> - add a focused scraping test target and downloader regressions
> 
> ## Why
> Remote text sources can be served with `Content-Encoding: gzip` or `deflate`. Passing those encoded bytes into extraction produces UTF-8 replacement characters and corrupts full-text and embedding indexes.
> 
> ## Verification
> - `zig fmt zig/lib/httpx/src/client/client.zig zig/lib/scraping/src/mod.zig zig/build.zig`
> - `zig build lib-scraping-test` passed before the follow-up deflate compatibility expansion
> - Draft CI is the compile/test gate for the final zlib/raw compatibility changes; a new local Zig build was intentionally not started because this checkout documents that local Antfly Zig builds may OOM.
> 
> Piledriver report: `.piledriver/compressed-chunks/report.md` in the Colony investigation workspace.

### DIFF
--- a/zig/build.zig
+++ b/zig/build.zig
@@ -2503,6 +2503,13 @@ pub fn build(b: *std.Build) void {
     const lib_regex_test_step = b.step("lib-regex-test", "Run standalone lib/regex tests");
     lib_regex_test_step.dependOn(&run_lib_regex_tests.step);
 
+    const lib_scraping_tests = b.addTest(.{
+        .root_module = scraping_mod,
+    });
+    const run_lib_scraping_tests = b.addRunArtifact(lib_scraping_tests);
+    const lib_scraping_test_step = b.step("lib-scraping-test", "Run standalone lib/scraping tests");
+    lib_scraping_test_step.dependOn(&run_lib_scraping_tests.step);
+
     const lib_jsonschema_tests = b.addTest(.{
         .root_module = jsonschema_mod,
     });
--- a/zig/build.zig
+++ b/zig/build.zig
@@ -2804,13 +2804,6 @@ pub fn build(b: *std.Build) void {
     const lib_embeddings_test_step = b.step("lib-embeddings-test", "Run standalone lib/embeddings tests");
     lib_embeddings_test_step.dependOn(&run_lib_embeddings_tests.step);
 
-    const lib_scraping_tests = b.addTest(.{
-        .root_module = scraping_mod,
-    });
-    const run_lib_scraping_tests = b.addRunArtifact(lib_scraping_tests);
-    const lib_scraping_test_step = b.step("lib-scraping-test", "Run standalone lib/scraping tests");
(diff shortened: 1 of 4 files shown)


ajroetker:
[codex]

I found three issues in the current revision (`eb816a7c6`); I would not merge it yet.

1. **P1 — Compressed writer responses spin indefinitely.**

   At `zig/lib/httpx/src/client/client.zig:2747`, the flate decompressor receives a `PrefixedReader`. That reader returns zero when flate requests data through its internal buffer, causing the loop at line 2750 to make no progress. Running `zig build lib-httpx-test` reproduced this as a 100% CPU loop. This affects gzip and both deflate forms through `getToWriter`.

2. **P1 — The new scraping tests deadlock before connecting.**

   At `zig/lib/scraping/src/mod.zig:1931`, the test supplies a security config whose default blocks private IPs, so the request to `127.0.0.1` is rejected. The error path then joins the server thread while it remains blocked in `accept`. Consequently, `zig build lib-scraping-test` hangs. The fixture should set `block_private_ips = false`, and error cleanup should wake or close the accepting server before joining.

3. **P2 — Two-byte sniffing cannot reliably distinguish zlib from raw deflate.**

   At `zig/lib/httpx/src/client/client.zig:2499`, any RFC 1950-looking prefix is classified as zlib, but a valid raw-deflate stream can begin with the same bytes. I constructed a valid raw stream beginning `78 9c`; raw decompression succeeds while zlib decompression fails. Preserving raw compatibility requires fallback after validation or buffering before choosing the decoder.

`git diff --check` passed. Both focused test commands were stopped after reproducing their respective hangs.

dovinmu:
> Addressed all three findings in `82b644fa5`:
> 
> 1. `PrefixedReader.readVec` now uses `Io.Reader.writableVector`, so flate can make forward progress through the adapter. The gzip, zlib-deflate, and raw-deflate writer cases remain covered.
> 2. The scraping fixture explicitly permits loopback and closes the listening server before joining its thread on request errors.
> 3. RFC 1950 header detection is now a hint rather than a final choice: buffered zlib decoding falls back to raw deflate after validation failure. Added a regression stream that is valid raw deflate, begins `78 9c`, and fails zlib decoding.
> 
> Verification on the updated head: [`zig-base / x86_64`](https://github.com/antflydb/antfly/actions/runs/31728400392/job/94542795315) passed, including the `lib-httpx-test` and `lib-scraping-test` dependencies in the unit-test build graph. `zig-base / arm64-codec` and the E2E executable build also passed; the two E2E runtime shards are still queued.

PR #492, opened 2026-08-16
Fix Raft apply conflict and managed embedding lifecycle flakes

ajroetker:
## Summary

- treat deterministic `IntentConflict`, `VersionConflict`, `DecisionConflict`, and `TxnNotFound` results as rejected Raft commands without stopping replica apply
- retain typed outcomes only for live synchronous leader proposals, binding each waiter to its exact log term and returning outcome-unknown for snapshot-covered or overwritten proposals
- preserve expected shard topology in single-index status responses and reject missing indexes before consulting runtime owners
- match same-name embedding index status by catalog incarnation and semantic config fingerprint
- selectively refresh the live writer only when the requested same-incarnation index has incomplete repair or coverage proof
- fence placement upserts and removals by metadata version and target drain state, retaining a per-placement version tombstone across delete/recreate
- use the existing placement transition commands and tags for the fenced shape; there is no legacy or v2 implementation
- reject placement mutation after node/store ownership is finalized
- overlay same-report live-writer facts onto conservative cached group status so hydrated relocation targets can converge without opening databases on the reporting path
- require query-fresh shard observations before managed-embedding lifecycle tests begin worker, retry, pacing, restart, or rebuild assertions
- reserve exclusive group and structural maintenance before draining live status observations, routing later status polls to the published snapshot so polling cannot starve recovery
- clear dense HBC state at actual restore-rebuild mutation boundaries while leaving proof-only worker-convergence retries free of cache and reader churn

## Invariants and UX

- deterministic command rejection advances the applied index on every replica while returning the exact typed result to the synchronous proposer
- a snapshot can prove applied progress but cannot fabricate a successful request outcome
- a higher-term entry at the same index cannot satisfy an older proposal waiter
- outcome memory is proportional to live requests, not Raft history; followers retain no request-scoped outcome state
- healthy table/list status paths retain their existing behavior, and healthy single-index reads avoid unnecessary writer refreshes
- same-name delete/recreate cannot report a stale incarnation as current
- incomplete same-incarnation status cannot mask a completed live-writer observation
- stale placement plans are deterministic no-ops after drain cancellation, a newer placement version, removal/recreation, or ownership finalization
- placement fences add metadata control-plane reads only; they do not add request-path work
- group-status overlay is scoped to the same group and `live_writer_publish` source, advances only conservative facts, and performs no DB opens
- known but unobserved shards remain visible as missing instead of collapsing into an empty `shard_status`
- `_coverage_incarnation` remains catalog-internal: public requests reject it and public index responses redact it
- status-only provisioned opens report persisted lifecycle state without performing writer-owned backfill work
- already-live status observations drain before exclusive mutation; later polls remain available from immutable published state and do not extend the drain
- foreground reads retain their stronger generation-lease fence, while proof-only restore retries do not invalidate readers or clear HBC state

## Validation

- focused Raft regression: 1/1 passed, covering all four deterministic results, continued apply progress, 5,000 concurrent cross-group waiters, snapshot coverage, and higher-term replacement
- placement CAS/fence, codec/projection, real control-loop, lifecycle hook, permit hook, status, workflow, and relocation retirement regressions passed
- index-status regressions passed 7/7 in both root test variants
- store-status cache regressions passed 5/5; focused live-writer overlay regression passed
- table-write subsystem passed 94/94 with no leaks; the two new `std.Io` concurrency regressions also passed under ThreadSanitizer
- production Antfly binary build, Zig format, and diff checks passed
- latest concurrent historical-flake soak: 30/30 passed across three workers, with 15 Autograph and 15 rate-limited managed-embedding recovery runs
- prior concurrent mixed-load E2E soak: 45/45 passed across three workers: 15 managed backup/restore, 15 delete/recreate rate-limit recovery, and 15 autoscaling drain/finalization runs
- prior concurrent low-FD soak: 60/60 passed across three workers, with 30 Autograph and 30 managed-embedding recovery runs
- lifecycle breadth pass: 8/8 tests using the query-fresh precondition passed


ajroetker:
[codex] Fixed both failures at their actual ownership boundaries in 52b7ba924.

- The x86 unit failure was deterministic: the strict public API parity manifest still named the pre-split index-status test. The manifest now includes both replacement tests, so neither incarnation-refresh case silently loses coverage.
- The autoscaling failure was generation starvation in the E2E polling shape. Each successful `/internal/v2/reallocate` request intentionally replaces the active barrier generation, but the wait loop submitted a new request every 500 ms. Under CI load, reconciliation could keep observing a superseded generation and never publish the balancing move. Add-node balancing now submits one generation and then polls state read-only. Drain and finalize waits submit none because those production mutations already trigger reconciliation themselves.

Validation:
- both renamed index-status tests passed through `public-api-parity-test`
- add-node placement passed 30/30 under 3 concurrent regression workers
- add, drain/replacement, shutdown/finalize, and node-churn scaling flows all passed on the final helper shape
- Ruff and `git diff --check` passed

This avoids timeout inflation and reduces metadata/API work while preserving the reallocation barrier invariant.

ajroetker:
[codex] Follow-up CI fixes are pushed in `4486cd605`.

The latest failures had two implementation causes plus one stale unit expectation:

- Single-index detail reads trusted a fresh cached observation even when the matching incarnation still advertised incomplete repair/coverage or active backfill. The detail path now performs a targeted writer refresh in that state and prefers the matching live-writer result. Healthy table/list requests and completed single-index cache hits keep the existing no-extra-writer-read path.
- A reconcile plan computed before placement removal could later recreate that placement on a draining node. Metadata apply now allows updates to an existing draining placement, but rejects creation after removal while the target node or store is draining.
- The provisioned-status unit expected a status-only DB open to execute writer-owned backfill. It now asserts the persisted lifecycle contract: backfill remains active and physical counts remain zero until a writer advances it.

Validation:
- focused status and metadata apply regressions: pass
- repository mixed-load regression script: 45/45 pass across 3 concurrent workers
  - 15 managed chunked backup/restore runs
  - 15 managed embedding delete/recreate + rate-limit recovery runs
  - 15 autoscaling drain/stop/finalization runs
- `zig fmt`, `git diff --check`, and the bounded production binary build: pass

The public API still strips `_coverage_incarnation`; the added response test covers that invariant.

ajroetker:
[codex] Follow-up review and fix pushed in b927de859.

Findings addressed:

- The requested-index detail path treated derived repair degradation as terminal, so a retained incomplete lifecycle snapshot could mask a completed live-writer observation after managed-index recovery.
- Placement transitions were vulnerable to stale reconciliation plans across drain cancellation, mutation, delete/recreate, and node finalization.
- Repeated E2E execution exposed a separate scaling race: a relocation target could publish a hydrated live writer while the same store report still carried a conservative empty cached group status, preventing promotion/finalization.

Implementation:

- Refresh only the requested same-incarnation index when its repair or coverage proof is incomplete; explicit issues/load failures and unrelated table/index debt keep the cached fast path.
- Make the existing placement upsert/remove commands and wire tags carry version/drain CAS preconditions directly. There is no legacy or v2 command path. A retained per-placement version tombstone fences ABA, and missing/mismatched node-store ownership rejects post-finalization mutation.
- Overlay only same-group `live_writer_publish` facts collected in the same store report onto conservative cached group facts before placement annotation. This adds no DB opens and cannot regress already observed facts.

Correctness/performance review:

- CAS failures are deterministic no-ops and the reconciler retries from a fresh snapshot.
- Extra reads are confined to metadata placement mutations, outside user request paths.
- Healthy index detail reads remain writer-free.
- Store-status reporting retains its cached/nonblocking behavior.

Validation:

- Placement fence/codec/projection/control-loop, lifecycle hook, permit hook, metadata status, workflow, and relocation retirement regressions passed.
- API index status: 7/7 passed in both root variants.
- Store-status cache: 5/5 passed; focused live-writer overlay test passed.
- Production Antfly build, format, and diff checks passed.
- Mixed repeated E2E: 45/45 passed, comprising 15 managed backup/restore, 15 managed-embedding delete/recreate rate-limit recovery, and 15 autoscaling drain/finalization runs.

ajroetker:
[codex] Follow-up review fixes are pushed in 4091245ba.

Findings addressed:

- Node finalization preflight only rejected active lifecycle records, so a stale safe-to-terminate observation could race remaining placement, membership, or runtime debt. The committed apply path also returned an error on a raced active node, which could prevent replicas from advancing the applied index.
- Live-writer data overlays were renewing the Raft group-status timestamp, making cached membership/readiness evidence appear fresh.
- Requested-index status trusted any retained live-writer snapshot, even when its lifecycle proof was incomplete.
- The focused flake run exposed a scheduler fairness bug: frequent immutable status reads could repeatedly beat the edge-triggered managed-index repair pass, leaving repair summary readiness degraded indefinitely.

Fixes:

- Finalization now uses the same placement/group/runtime debt fence as shutdown status, with a deterministic apply-time no-op if debt reappears after preflight.
- Live overlays accept only fresh live-writer reports, advance only monotonic data facts, and do not renew the Raft freshness fence.
- Requested-index status performs a targeted writer refresh for incomplete retained snapshots, while complete and explicit terminal states remain cache-only.
- Index repair setup now uses the existing read-compatible operation gate. Queries/status continue from immutable published generations, conflicting mutations remain excluded, and polling can no longer starve repair. Ordinary startup/restore catch-up stays exclusive.
- The 409 response now says the node is not ready to finalize, which covers all lifecycle debt rather than incorrectly implying only an active-node conflict.

Validation:

- Focused metadata apply/finalization, runtime overlay, API index-status, and repair-admission unit tests passed.
- Bounded production antfly build passed.
- Exact managed delete/recreate rate-limit reproduction: 30/30 passed across three workers.
- Broader managed semantic backup/restore + index lifecycle + autoscaling drain/finalize matrix: 45/45 passed.

ajroetker:
[codex] Follow-up review fixes are pushed in ea4cb9c07.

The review found and fixed five correctness gaps:

- Read-compatible index-repair admission was too broad: a cold owner may perform restore/replay/artifact rebuild work that must remain read-exclusive. Repair now uses read-compatible admission only while holding a live cached-writer lease; otherwise it hands off to the ordinary exclusive lifecycle guard. The extra lease closes the observe-then-retire TOCTOU.
- A committed finalize command could become a deterministic no-op when new debt appeared after preflight, silently losing the accepted one-shot request. Finalization now persists a finalizing lifecycle and is retried transactionally by only the store/placement transition capable of clearing that node debt. Cancellation still wins if it commits first.
- Shutdown status filtered foreign runtime observations, while preflight/apply counted every retained observation. All three paths now share one identity-aware termination-debt predicate, so safe_to_terminate exactly matches the state-machine deletion fence.
- The artifact-repair summary worker existed but was never started for managed cached writers, so a repaired index could remain repair_summary_ready=false until an unrelated reopen. It now starts only after the DB is installed at its stable cache address; read-only and short-lived catch-up owners remain worker-free.
- A RuntimeStatusPublicationFenced result could abort repair after durable mutation and strand the intent in rebuilding. The stale snapshot is still rejected, but runtime-status publication is now explicitly observational and cannot veto durable maintenance progress; non-fencing errors still propagate.

Invariant/performance notes:

- Placement ownership and matching store debt remain hard deletion fences.
- Cold recovery stays read-exclusive; only a pinned live writer mutating an unpublished candidate generation is read-compatible.
- Pending finalization uses targeted point lookup on normal store updates and scans placement/store state only for the rare finalizing lifecycle; there is no metadata-wide poller.
- Artifact-summary maintenance is bounded and sleeps while idle, and it runs only on the long-lived writer owner.
- No legacy compatibility path or new public API field was added. finalizing is an internal lifecycle value surfaced through the existing shutdown phase.

Validation:

- Production antfly build passed via the bounded build script.
- Focused table-writes status-fence/admission tests passed (the aggregate root-test wrapper also reports its known unmatched sibling target).
- Focused metadata shutdown/preflight/Raft-apply tests passed.
- Exact managed rate-limit delete/recreate regression: 30/30 passed with 3 concurrent workers after the final fix.
- Final mixed regression matrix: 45/45 passed with 3 concurrent workers across managed backup/restore, rate-limited delete/recreate, and autoscaling drain/finalize.

PR #499, opened 2026-08-18
fix(operator): reject unsafe metadata replica changes

### LINKED ISSUE #498: operator: metadata scale-up 1→3 leaves split incarnations while cluster reports Available
## Summary

With the v0.2.0 runtime and the latest published operator, changing a distributed `AntflyCluster` from one metadata replica to three on retained PVCs leaves the original member in its old one-voter incarnation while the two new members report a different three-voter incarnation. All three pods are Ready and the operator reports the cluster Running/Available.

This creates a green split control plane: callers can be routed to metadata members with divergent catalogs and leaders.

## Versions

- Runtime: `ghcr.io/antflydb/antfly:v0.2.0@sha256:654ebfd7360b8f48e964b24e674e5e5236e57cafc1760c66e6f3718ec15fad1e`
- Operator: `ghcr.io/antflydb/antfly-operator:v0.0.17-rc.8@sha256:b68c8d7d5e9ae1855eb2619990e273e0aaf1e99d9fb8a86b33444f27fc50ac48`
- Kubernetes: EKS, all runtime pods on arm64
- Distributed topology: three data replicas throughout; metadata changed from one to three

## Reproduction

1. Create a distributed `AntflyCluster` with `metadataNodes.replicas: 1` and persistent metadata storage.
2. Wait for Running/Available and the metadata StatefulSet to become Ready.
3. Update the CR to `metadataNodes.replicas: 3`, retaining the existing PVC.
4. Wait for the operator to report `metadataNodesReady: 3`, Running/Available, and for all three StatefulSet pods to become Ready.
5. Query `/metadata/v1/status` directly on each metadata pod.

The state remained divergent after repeated polling for more than five minutes:

| Pod | Incarnation | Local node | Role | Leader | Term | Commit | Voters | Lease owner | Holds lease |
| --- | --- | ---: | --- | ---: | ---: | ---: | ---: | ---: | --- |
| metadata-0 | `40941dd63c7a6e051cd67f07d72b35e9` | 1 | leader | 1 | 3 | 9021 | 1 | 1 | false |
| metadata-1 | `34e1d255e8a5b74b113d4c5cffdb22e6` | 2 | follower | 3 | 2 | 6120 | 3 | 3 | false |
| metadata-2 | `34e1d255e8a5b74b113d4c5cffdb22e6` | 3 | leader | 3 | 2 | 6125 | 3 | 3 | true |

All three reported the same `metadata_group_id`, which makes a group-ID-only health check insufficient.

## Expected behavior

Either:

1. metadata scale-up joins the new members to the existing incarnation through a quorum-aware membership transition, or
2. the admission webhook rejects metadata replica-count changes and tells the operator to recreate the cluster at the target topology.

In both cases, Running/Available and MetadataReady should not be true while members disagree on incarnation, leader, and voter configuration.

## Notes

The update webhook currently rejects metadata scale-down but permits scale-up. The generated StatefulSet updates every member to the new three-peer `--cluster` argument, but the retained ordinal-0 Raft state does not adopt the new incarnation/membership. A cluster created from empty PVCs at the target three-member topology is being tested separately.

ajroetker:
Summary:
- reject metadata replica-count scale-up and scale-down in the validating webhook and controller fallback
- keep the generated base CRD compatible with Kubernetes 1.23-1.24, with an optional Kubernetes 1.25+ Kustomize overlay for CEL transition validation
- durably record and monitor metadata topology without catalog-sized steady-state work
- stop reconciliation if the topology status checkpoint conflicts, preventing a stale replica count from becoming immutable
- update operator documentation and regression coverage

Testing:
- KUBEBUILDER_ASSETS=<1.35 assets> GOWORK=off GOEXPERIMENT=simd go test ./...
- GOWORK=off GOEXPERIMENT=simd go vet ./...
- kubectl kustomize manifests/overlays/kubernetes-1.25

Fixes #498


excenter:
<3 ty

PR #505, opened 2026-08-18
perf(algebraic): reduce exact cardinality overhead

ajroetker:
## What changed

- uses hash-set membership for exact doc-id constraints
- makes child cardinality evaluation single-pass
- removes row scaffolding and unused helpers
- accumulates additive results natively

## Why

These PR #141 optimizations are result-preserving and independent of HLL, adaptive materialization, relational storage, and the larger #145 integration.

## Impact

Exact algebraic cardinality paths allocate less and avoid repeated traversal while retaining canonical results.

## Validation

- 3 focused algebraic cardinality/fold/distributed-merge tests passed, no leaks
- `git diff --check`

### DIFF
--- a/zig/pkg/antfly/src/storage/db/algebraic/index.zig
+++ b/zig/pkg/antfly/src/storage/db/algebraic/index.zig
@@ -4777,8 +4777,10 @@ pub const Index = struct {
             buckets.deinit(self.alloc);
         }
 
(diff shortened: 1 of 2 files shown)


ajroetker:
Superseded by #509, which consolidates the extracted cardinality and dynamic-template algebraic work into one PR.

PR #512, opened 2026-08-19
Optimize public vector ingest memory and compaction

ajroetker:
## Summary

- remove whole-mutable-state cloning from point probes, erased storage adapters, and transaction intent resolution
- batch dense identity, mapping, split-metadata, and transformed-vector reads; reuse split materialization and routing state
- bound live dense replay windows and give replay, compaction, and foreground writes explicit pressure/fairness behavior
- fix public write-coalescer ownership starvation and use the ordinary atomic shard path for one-participant batches
- preserve hard L0 pressure during continuously replenished primary/dense bulk windows
- allow unlocked compaction output to publish after concurrent L0 prepends by relocating exact immutable input IDs and recomputing the current target overlap closure
- raise macOS build scheduler reservations to observed compile peaks while retaining normal parallel builds
- document the investigation and accepted/rejected A/B results in `zig/VECTORDBBENCH_FINDINGS.md`

The unlocked-compaction issue was the final large throughput failure. Persisted output is built without the backend lock. Concurrent flushes prepend newer L0 runs, so positional validation discarded valid completed compactions even though every selected immutable run still existed. The new publication path accepts only a pure input-ID relocation; it still rejects a removed input or changed target-level overlap closure.

## Benchmark contract

These are upstream VectorDBBench loads through Antfly's public `/db/v1` server and table/index lifecycle, one shard, batch 100, four load workers, packed vectors, and `sync_level=write`. Readiness is polled through the public dense-index status. The default full-text index is removed through the public API by the companion VectorDBBench change so unrelated asynchronous full-text work is not charged to the ANN case.

| Workload | Insert | Ready total | Demand | Peak RSS | Final L0 |
| --- | ---: | ---: | ---: | ---: | ---: |
| OpenAI 50K | 25.05 s | 27.53 s | 708.5 MB | 1.96 GB | 59 / zero debt |
| Cohere 1M before relocated publication | 1,536.04 s | 1,644.37 s | 1.60 GB | 3.89 GB | 825 / debt 582 |
| Cohere 1M after relocated publication | 1,099.85 s | 1,110.66 s | 1.20 GB | 4.21 GB | 111 / zero debt |

The fixed 1M run published 24 compactions from 24 pressure events with zero overloads, versus 16 publications from 26 events before the fix. It was 32.5% faster end-to-end than the preceding full run and 63% faster than the reported roughly 3,000-second run. The host remained available to unrelated work, so 18.51 minutes is a contended diagnostic rather than a publication-grade mean or a claim that the hoped-for roughly 13-minute target is fully reached.

## Validation

- `zig build lsm-backend-test -Doptimize=ReleaseFast --summary all`: 303 passed, 1 platform helper skipped, 0 failed/leaked
- `zig build -Doptimize=ReleaseFast --summary all`: 24/24 steps passed with normal parallel scheduling
- upstream public-API VectorDBBench OpenAI 50K: 50,000 vectors query-visible
- upstream public-API VectorDBBench Cohere 1M: 1,000,000 vectors query-visible, no request timeout

## Companion PRs

- VectorDBBench ANN/full-text isolation: https://github.com/antflydb/VectorDBBench/pull/13
- Circus low-overhead footprint sampler: https://github.com/antflydb/antfly-circus/pull/52


bpopadiuk (commented):
Codex (GPT-5):

No blocking findings at reviewed head `030b79c94cd6f77bcf6198d77b261908df94a23e`.

Local validation:
- `zig build lsm-backend-test -Doptimize=ReleaseFast --summary all`: 303 passed, 1 platform skip.
- Focused transaction probe regression: passed.
- `zig build lib-vectorindex-test -Doptimize=ReleaseFast --summary all`: passed.
- `git diff --check`: clean.

CI is not yet complete: `zig-base / arm64-codec` passed; `zig-base / x86_64` remains queued and `e2e-base / build` is in progress.

PR #519, opened 2026-08-20
Label controlled release archive downloads

bpopadiuk:
Implemented by Codex.

Summary:
- send User-Agent antfly-installer/1 on installer archive requests
- accept ANTFLY_DOWNLOAD_CLASS=external, employee, or ci; only employee and ci send the metric-label header
- treat missing or invalid values as external without breaking installs
- label direct CLI packaging downloads as release automation and CI
- reject redirects for labeled installer and release-automation requests so metric headers remain on the versioned release host
- add a hermetic installer and release-workflow marker test

These headers are caller-asserted metric labels only. They can be omitted or forged and must not authorize access or prove employee or CI identity.

Validation:
- installer and release-workflow marker test passes
- sh/bash syntax checks pass
- ShellCheck passes (excluding the installer script's pre-existing POSIX local warning)
- actionlint passes
- a live one-byte versioned archive request returned 206 with Content-Range and zero redirects while redirect rejection was enabled
- git diff --check passes

### DIFF
--- a/.github/workflows/cli-publish.yml
+++ b/.github/workflows/cli-publish.yml
@@ -55,6 +55,7 @@ jobs:
             "antfly_${version}_Linux_x86_64.tar.gz"
           do
             curl -fsSL \
+              --max-redirs 0 \
               -H "X-Antfly-Download-Channel: release-automation" \
               -H "X-Antfly-Audience: ci" \
               "${base}/${archive}" \
--- a/.github/workflows/cli-publish.yml
+++ b/.github/workflows/cli-publish.yml
@@ -54,7 +54,11 @@ jobs:
             "antfly_${version}_Linux_arm64.tar.gz" \
             "antfly_${version}_Linux_x86_64.tar.gz"
           do
-            curl -fsSL "${base}/${archive}" -o "dist/release-archives/${archive}"
+            curl -fsSL \
+              -H "X-Antfly-Download-Channel: release-automation" \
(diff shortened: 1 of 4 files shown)


ajroetker:
One requested CI follow-up: the new `scripts/test_install_download_markers.sh` is not currently invoked by any workflow, and on this PR only the `changes` job ran while the substantive jobs were skipped.

I suggest using `.github/workflows/zig-tests.yml` as the repository-wide PR entry point, adding a `release_tooling` output to its existing `changes` classifier for:

- `scripts/install.sh`
- `scripts/test_install_download_markers.sh`
- `.github/workflows/cli-publish.yml`
- `.github/workflows/zig-tests.yml`

Then add a small path-gated `release-tooling` job that checks out the repo and runs:

```sh
scripts/test_install_download_markers.sh
sh -n scripts/install.sh
bash -n scripts/test_install_download_markers.sh
```

ShellCheck/actionlint can be included if they are already available or cheap to set up, but the behavioral marker test is the important PR gate. I would not put this in `cli-publish.yml`, since that workflow runs during publishing/dispatch/tag events and attempts to download an already-published release rather than serving as a PR check.

I ran the marker test, syntax checks, ShellCheck, actionlint, and `git diff --check` locally against the PR head; all passed.

ajroetker:
gonna cancel the unnecessary e2e run to free up the runner

PR #559, opened 2026-08-25
inference: require configured execution backend

bpopadiuk:
## Summary

- add `ANTFLY_INFERENCE_REQUIRED_BACKEND` as a fail-closed runtime policy distinct from backend preference
- restrict model loading, direct ONNX paths, and compiled/artifact selection to the required backend when configured
- reject invalid values, distinguish an unavailable backend from a backend-specific load failure, and preserve the policy across server/model-manager copies
- keep existing automatic fallback behavior unchanged when no required backend is configured

## Motivation

GPU-hosted inference must not silently fall back to CPU when CUDA is unavailable. This allows production deployments to require CUDA while preserving the existing preferred-backend and test-harness behavior.

## Out of scope

- provisioning GPU infrastructure
- publishing or selecting a production container image
- choosing the production model bundle
- proving execution on a real CUDA device

## Validation

- focused required-backend suite: 9 selected, 9 passed
- full `zig build inference-test`: 3,156 selected; 3,138 passed; 18 skipped
- `git diff --check`

### DIFF
--- a/zig/pkg/inference/src/backends/backends.zig
+++ b/zig/pkg/inference/src/backends/backends.zig
@@ -360,8 +360,17 @@ pub const SessionManager = struct {
     /// different valid required backend makes the fast path ineligible.
     pub fn allowsDirectBackend(self: *const SessionManager, backend: BackendType) !bool {
         if (self.required_backend_invalid) return error.InvalidRequiredBackend;
+        if (self.required_backend) |required| {
+            if (!required.supportsDirectSessionLoad()) return error.RequiredBackendUnavailable;
+        }
+        return self.allowsBackend(backend);
+    }
+
+    /// Whether a backend-specific execution path is compatible with the
+    /// process-level required backend, including compiled artifact runtimes.
+    pub fn allowsBackend(self: *const SessionManager, backend: BackendType) !bool {
+        if (self.required_backend_invalid) return error.InvalidRequiredBackend;
         const required = self.required_backend orelse return true;
-        if (!required.supportsDirectSessionLoad()) return error.RequiredBackendUnavailable;
         return required == backend;
     }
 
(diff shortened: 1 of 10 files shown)


ajroetker:
[codex] Review findings

1. **P1 — The fail-closed policy is bypassed by public compiled-artifact entrypoints.** The new validation helper is only called by `generate` and the server. `run-artifact` proceeds directly to `runArtifactPrompt`, and ONNX artifacts create sessions directly; likewise, `smoke --backend onnx|xla` assigns the compiled backend without validating it. Consequently, `ANTFLY_INFERENCE_REQUIRED_BACKEND=cuda` can still execute an ONNX/PJRT artifact through these commands. The check should live in the shared artifact execution path and also guard the smoke pipeline.

2. **P2 — `pjrt` is accepted as a required backend but cannot reach its compiled execution path.** `allowsDirectBackend()` returns `RequiredBackendUnavailable` whenever the required backend lacks direct-session support; both native generate and the server call it unconditionally before compiled selection. Thus `required=pjrt` rejects even `backend=xla`, despite `validateRequiredCompiledBackend` explicitly allowing that combination. Either PJRT should cease being accepted as a required policy value, or compiled-only required backends need a path that does not first demand direct session loading.

3. **P3 — The new compiled-backend policy test is not part of `inference-test`.** Running `zig build inference-test -fincremental -- --test-filter 'compiled backend execution'` reports “No tests matched runtime filters.” The focused required-backend command still reports 9/9 because it omits the `native_backend_choice.zig` test. Please add that module directly to the root test imports.

The worktree was clean at `9d2ca7b40`; focused required-backend tests passed 9/9, and current base CI checks pass. I would request changes for the artifact-policy bypass.

PR #560, opened 2026-08-26
fix(jpeg2000): stream native irreversible decoding

dovinmu:
## Summary

- Replace the macOS-only JPEG 2000 fallback heuristic with a bounded pure-Zig decode path.
- Decode Tier-1 codeblocks one at a time and write them directly into component planes, releasing each coefficient grid before the next codeblock.
- Run inverse wavelet reconstruction directly on strided component regions and compact reconstruction storage before allocating interleaved output.
- Honor effective per-component COC wavelet overrides; mixed transforms stream when safe and retain the established mixed/MCT path otherwise.
- Cover both the original 7.54 MP OfficeQA JPX image and a 9.31 MP residual-failure image under enforced native memory ceilings.

## Root cause

The OfficeQA Treasury PDFs use high-resolution, three-component irreversible JPEG 2000 images. The previous decoder retained every Tier-1 coefficient grid, all floating-point reconstruction planes, full-resolution DWT scratch, and duplicate integer/output planes concurrently. PR head `ed22e834` reduced the original 1985 fixture below 96 MiB, but three full-resolution component planes alone require roughly 12 bytes per pixel. The residual cohort contains images up to about 9.7 MP, whose irreducible plane storage exceeds 96 MiB before bounded bookkeeping.

The final change keeps the streaming decoder and raises the PDF decoder ceiling to 128 MiB. This remains below the existing 256 MiB document-extraction slice hard limit and covers the measured residual geometry without an unbounded or platform-specific fallback.

## Memory and correctness

- Original `2319x3253` fixture: enforced at **96 MiB**
- Larger `2612x3564` fixture: enforced at **128 MiB**
- Original decoded pixel SHA-256 remains `bd592833fd9c09b0501d9eb6855c50e557d4df9afaca7d34963edb5af5c1363f`
- Both source codestreams are checked in with provenance and source SHA-256.

## Verification

- Required direct decoder artifact: `zig build lib-image-jpeg2000-test --summary all` — **247/247 passed**
- `zig build lib-image-test --summary all` runs both the wrapper and direct decoder artifacts
- `zig build lib-pdf-test --summary all` — **220/220 passed**
- `zig build lib-pdf-safety-test --summary all` — **3/3 passed**
- Bounded ReleaseFast Antfly build completed
- Full 150 DPI render passes completed for the five largest representative residual failures
- Product ingestion with OCR selection enabled:
  - `1982_03`: ready/fresh, 830 derived documents
  - `1972_05`: ready/fresh, 678 derived documents
  - `1976_04`: ready/fresh, 659 derived documents
  - `1970_01`: ready/fresh, 711 derived documents
  - `1973_05`: ready/fresh, 718 derived documents
- Full residual cohort sweep from the `ed22e834` evidence set: **77/77 complete**, zero `DocumentExtractionWorkingSetTooLarge`

## Baseline findings

- The external OpenJPEG suite's `p0_13` tolerance failure reproduces unchanged on the parent commit.
- The local aggregate image conformance step also encounters an existing WebP corpus failure; focused image, PDF, JPEG 2000, and required CI targets are green.

## Concurrent branch update

Commit `738dd4593` from the repair-route status-fencing work is retained on this branch at Rowan's direction. PR #565 remains the canonical full repair-handoff change.


zig/lib/pdf/src/reader.zig
```diff
@@ -3531,6 +3543,12 @@ pub const Reader = struct {
         const pixel_count = std.math.mul(usize, width, height) catch return error.UnsupportedPdfRendering;
         if (pixel_count > 100_000_000) return error.RenderedPageTooLarge;
         const rgba_len = std.math.mul(usize, pixel_count, 4) catch return error.UnsupportedPdfRendering;
+        // The pure Zig JPEG 2000 path retains coefficient, wavelet, sample,
+        // and output planes concurrently. Defer large images to the platform
+        // renderer before allocator exhaustion so the caller's normal
+        // UnsupportedNativeDecode fallback remains available.
+        if (has_jpx and jpeg2000NativeDecodeExceedsWorkingSet(rgba_len, self.decode_limits.max_working_set_bytes))
```
ajroetker: We don't build with the platform decoder for native builds so we need a proper implementation

dovinmu: Agreed. The platform fallback heuristic has been removed in f6ba46f77. The replacement is pure native: Tier-1 codeblocks are decoded and consumed one at a time, inverse 9/7 DWT operates directly on strided component regions, and f32 reconstruction buffers transfer to i32 ownership in place. On the exact 2319x3253 JPX stream, tracked peak live allocation fell from 295,755,317 to 114,162,517 bytes with identical decoded pixel SHA-256. A native-only 128 MiB regression test now enforces that ceiling, and the full 238-page PDF renders without relying on UnsupportedNativeDecode fallback. Ready for another look.

ajroetker:
[codex] Fixed the underlying native decoder issue rather than relying on the platform-renderer fallback.

The JPEG 2000 path now streams Tier-1 codeblocks into reconstruction planes, performs inverse wavelet reconstruction in place, and compacts one component plane before allocating the interleaved RGB output. This keeps the exact 2319×3253 OfficeQA fixture within the original 96 MiB reservation without raising the configured limit.

Measured peak live decode memory:
- Before final compaction: 114,162,517 bytes
- After: 91,845,631 bytes

The regression fixture produces the same SHA-256-verified pixels. Validation passed:
- 245 direct JPEG 2000 tests
- 220 PDF tests
- ReleaseFast OfficeQA regression test

ajroetker:
transient, ill fix in another pr

ajroetker:
[codex] Review findings:

1. **Mixed component wavelet transforms are routed through the wrong reconstruction path.** `decodeU8Bytes` selects streaming from the default COD transform only, but COC can override the transform per component. `IrreversiblePlaneAssembler.finish` then applies inverse 9/7 to every component. Please stream only when every effective component coding style is irreversible; otherwise use the existing mixed-component reconstruction path. Add coverage for COD=9/7 with a COC=5/3 override.

2. **The OfficeQA memory regression is not part of required CI.** The test passes when `decode.zig` is tested directly, but `zig build lib-image-test` runs only the importing wrapper test (`1/1`), so the test at `decode.zig` is excluded. Please add a focused test artifact rooted at `decode.zig` (or equivalent) and make `unit-test` depend on it.

Non-blocking: the PR description still says 128 MiB, while the committed test enforces 96 MiB.

Local verification: direct JPEG 2000 suite 245/245; PDF suite 220/220; OfficeQA focused test passed; diff check clean. The current x86 failure is in the unrelated `api.table_writes.test.structural reconcile publishes durable index repair debt once per group` test; I am handling that separately from main.

ajroetker:
[codex] Follow-up on the pre-existing x86 failure: the failing test was structural reconcile publishes durable index repair debt once per group, at the assertion that status must no longer be snapshot-only after repair. The durable index repair itself completed. The bug was in the subsequent lifecycle handoff: both the repair-clear callback status publication and the final consistent status publication could be fenced by a table-epoch invalidation. RuntimeStatusPublicationFenced was intentionally swallowed as observational, but the result then reported no remaining repair, so the scheduler removed the exact group route and left snapshot-only authority stranded with no later publisher.

I reproduced that sequence deterministically and opened #565 with the underlying fix. It records the durable clear proof and keeps the exact repair route pending until a later authoritative publication settles the handoff. The focused 97-test production regression target passes (95 passed, 2 sandbox-only listener skips), plus 32/32 parallel repetitions of the exact race regression.

dovinmu:
[codex]
Addressed both review findings, plus the residual OfficeQA cohort:

1. Mixed COD/COC wavelet styles now route through `StreamingPlaneAssembler.canAssemble`. Effective per-component styles are honored; the streaming assembler stores irreversible components as `f32` and reversible components as `i32`, while mixed transforms with MCT retain the established reconstruction path. The direct suite includes a COD=9/7, COC=5/3 regression.
2. `lib-image-jpeg2000-test` is now a direct test artifact rooted at `decode.zig`. Both `lib-image-test` and required `unit-test` depend on it. Local result: 247/247 direct decoder tests passed; the aggregate reports 248/248.
3. The 96 MiB ceiling is retained for the original fixture, while the document decoder default is 128 MiB for the larger three-component images. A checked-in 2612x3564 residual fixture enforces the 128 MiB bound.
4. The exact 77 PDFs that remained broken under `ed22e834` were swept against the new ReleaseFast binary: 77/77 completed with ready/fresh derived indexes and zero `DocumentExtractionWorkingSetTooLarge`. The five largest representatives also passed individual product-ingestion checks.

The PR description now contains the full evidence and fixture provenance.

PR #561, opened 2026-08-26
fix: make provisioned restore and Raft replay converge

excenter:
## Summary

This is the upstreamable patch set from a real 3-metadata/3-data Kubernetes restore failure.

- make native restore use one prepare/drain/promote/publish generation handoff, including exact-content deduplication while readers are resident
- persist restore retry backoff and make CLI polling distinguish transient, ambiguous, and terminal HTTP outcomes
- make metadata table forwarding at-most-once, reject followers before parsing, validate restore progress before Raft admission, and enforce route body limits during H1/H2 transport admission
- make Raft replay identity independent of Ready-window batch boundaries, persist retry checkpoints, restore committed membership, and clear per-group apply state on permanent retirement
- advertise stable data-pod DNS from the operator and preserve inference registry variants during pull/preload
- stabilize C API quarantine-worker ownership

## Failure reproduced

A 3+3 provisioned cluster completed restore on one path, then Raft bootstrap repeatedly returned `GenerationTransitionActive` because a resident reader held the published generation. Other tests exposed replay failures when restart changed Ready batching and stale retry state after group retirement.

The new tests first reproduced those failures locally. The fixes move ownership and identity checks to the shared lifecycle/storage boundaries instead of adding call-site retries.

## Validation

Final combined run on this exact branch:

- restore CLI polling: 2/2
- restore retry/backoff: 4/4
- restore generation/bootstrap: 13/13
- data runtime: 122/122
- durable replay categorical cases: 5/5
- table forwarding: 6/6
- restore-progress admission: 6/6
- metadata service: 45/45
- metadata logic: 243/243
- Raft simulation: 10/10
- seeded metadata VOPR: 1/1
- storage authority: 76 passed, 1 environment-dependent S3 test skipped
- API HTTP/runtime: 43/43
- full httpx suite: passed
- Go operator `go test ./...` with `GOEXPERIMENT=simd`: passed
- formatting and staged diff checks: passed

The existing `lib-metadata-sim-core-test` automatic split/merge admission failures reproduce unchanged on untouched `main`; the focused Raft simulation and seeded VOPR suites above pass.

## Deployment state

Local and simulation proof is complete. A fresh immutable ARM64 image build and the full SalesHub 3+3 backup/drop/restore rehearsal are running separately; this PR remains draft until that live proof is attached.


ajroetker:
[codex] The core `db.zig` change makes sense and fixes a real lifetime/address-stability bug, but I do not think the implementation is complete yet.

Previously, `DB.open()` started the quarantine retry thread with a `*DB` pointing at its local `db` value and then returned that `DB` by value. Returning the struct can move/copy it, so the worker can retain the address of the temporary inside `DB.open()` rather than the address ultimately owned by the caller. Removing startup from `DB.open()` and starting only after the DB has been installed in stable storage is therefore the right direction.

The new C API placements look correct: each call occurs after the allocator-owned `Handle` has been initialized, so `handle.db` remains at a stable address until close. The managed table cache follows the same pattern by starting the worker only after installing the DB in its heap-allocated cache entry. The idempotence and writer/runtime guards in `startQuarantineRetryWorkerIfNeeded()` are also sensible, and the pointer-address regression assertions are useful coverage for this particular failure mode.

My concern is that moving startup out of `DB.open()` makes every production owner responsible for starting the worker, but the calls added by this PR appear limited to the C API and managed table cache. In particular, the embedded constructors in `zig/pkg/antfly/src/embedded/db.zig` still return their wrapper by value and do not start the quarantine retry worker. Those writer DBs previously received automatic background quarantine recovery through `DB.open()` and now appear to lose it silently.

Adding the call inside `embedded.DB.openWithProfile()` or `openWithLiteBackend()` is not sufficient on its own, because those functions also return the wrapper by value and would recreate the same address-stability problem.

Please audit every production owner/constructor of `storage.db.DB` and preserve the previous retry behavior. Some possible designs are:

- Make each long-lived DB owner address-stable before explicitly starting background workers, with tests covering every supported entry point.
- Introduce an explicit pin/finalize/start-background-workers lifecycle that cannot be skipped accidentally.
- Move the retry worker context into separately allocated stable storage so the worker does not retain a pointer to a movable `DB` value.
- Make DB ownership pointer-based/non-movable, if that larger ownership change is appropriate.

Until that audit is complete, I would consider this a substantive availability regression: quarantined indexes opened through an uncovered owner may no longer self-heal in the background and may require reopening or explicit operator action. I would classify it as P1 if automatic recovery is part of the availability contract, otherwise P2.

This worker-lifetime/ownership correction would also be easier to review as a focused PR or commit series separate from the broader restore/Raft convergence work. It establishes a subtle ownership contract across constructors and deserves its own call-site audit and regression coverage. A useful split would be:

1. The address-stability fix and ownership API, including coverage for all C API, managed, and embedded/native entry points.
2. The provisioned restore/Raft replay convergence changes.
3. Any independent preload or generated-code changes.

That split is not merely about PR size: it lets reviewers verify that the lifetime fix preserves behavior everywhere before evaluating the higher-level convergence changes that depend on it.

ajroetker:
Thanks @excenter for the contribution!! Boris said you had one incoming, I took a look through and had a look through with my codex and had codex post the summary of my review/convo there, it'd be nice to split up the pr into the independent subsystems (operator, parser, config, c api, backend db bug, raft) so they can be reviewed independently (codex missed things when it reviewed all together but found things when i broke it down by subsystem), sorry for the inconvenience but it touches some pretty sensitive parts of the codebase as you found and I'd like to make sure to add simulation/test coverage where you found issues as well

ajroetker:
[codex] Follow-up with the remainder of my review context (the quarantine-worker ownership finding is in the preceding comment).

## Resolved at the current head: preload registry variants

The earlier preload parsing behavior treated the first colon-delimited component after the model kind as a backend. That conflicts with registry references and artifact variants that legitimately contain colons. For example, a value such as `embedder:owner/model:i8` must preserve `owner/model:i8` as the model name, while `generator:metal:owner/model:i8` should recognize `metal` as an explicit backend and preserve the remainder as the name.

Without that distinction, eager preloads produced by the operator could be rejected or resolve a different model than the init container pulled. This was particularly relevant because this PR deliberately preserves references such as `:i8` and multi-component artifact variants.

At the current head, I consider this finding resolved:

- `8463513fe` changes the standalone parser to consume a backend component only when it is a recognized backend and adds CLI coverage for both registry variants and explicit backends.
- `ed88eff03` extracts the shared Antfly-side parser into `common/preload_model_spec.zig`, uses it from both standalone and inference-runtime entry points, and adds focused tests for `:i8`, multi-component `:gguf:Q4_K_M`, explicit backends, and malformed values.
- `zig/pkg/inference/src/main.zig` has equivalent behavior and focused coverage as well.

The shared parser is a worthwhile improvement because it removes the most immediate source of drift between the two Antfly command paths. The standalone inference executable still has a local equivalent parser, but its behavior and tests currently agree, so I do not see a remaining correctness blocker there.

## Remaining confirmed finding

The remaining actionable correctness concern is the quarantine retry worker ownership/call-site audit described in the previous `[codex]` comment. The underlying move-after-spawn fix is valid, but embedded/native owners appear to lose automatic retry startup.

## PR scope and suggested split

This PR currently changes 51 files with roughly 7.2k additions across several independently reviewable systems: operator DNS/routing, inference model-reference handling, C API/background-worker ownership, HTTP admission/body limits, restore generation transitions, metadata forwarding, and Raft replay/checkpoint semantics.

I recommend asking for a split, preferably while keeping the current branch as an integration branch if that helps preserve the live 3+3 validation. A practical dependency-aware series would be:

1. Operator stable data-pod routes.
2. Inference registry/preload parsing and tests.
3. Quarantine-worker address stability and complete owner coverage.
4. HTTP transport admission/body-limit changes.
5. Raft replay identity, retry checkpoints, membership restoration, and retirement cleanup.
6. Restore generation handoff, metadata forwarding/admission, and CLI retry semantics, based on the prerequisite commits above.

The split is important for correctness, not just reviewer convenience. The worker change establishes a pointer-lifetime contract across constructors; the HTTP changes affect shared H1/H2 transport behavior; and the replay/restore changes alter separate persistence and distributed-state invariants. Each area needs a reviewer to be able to reason about its contract and tests without simultaneously validating unrelated subsystems.

If splitting the branch would invalidate or make the live cluster proof difficult to reproduce, an acceptable compromise would be to keep this PR as the tested integration branch but open stacked prerequisite PRs for review and merge them in order, then reduce this PR to the final restore/convergence delta.

Aside from the resolved preload issue and the open worker-ownership concern, I do not have another confirmed line-level correctness defect to report from this pass. The breadth of the restore/Raft/HTTP changes is itself the main residual review risk and the reason I recommend decomposition before approval.

ajroetker:
[codex] Follow-up on the independently landable fixes found while reviewing this PR: I extracted them into five focused PRs against `main` so they can be reviewed, tested, and merged without waiting for the larger distributed-inference/state-machine discussion.

- #566 — stable operator data-pod routes. Keeps listeners bound to the Pod IP while advertising stable StatefulSet DNS names for API and Raft. The corresponding operator routing change can be dropped from this PR after merge. Validated with `go test ./controllers/antfly`.
- #567 — Bedrock request correctness. Fixes wall-clock/SigV4 timestamps, calendar month formatting, canonical URI percent escaping, and Titan multimodal output-dimension encoding. The corresponding Bedrock change can be dropped after merge. Validated with the focused Bedrock test target.
- #568 — restore polling resilience. Preserves non-2xx responses long enough to classify them, retries explicitly transient statuses, and gives ambiguous 404/500 responses a bounded **consecutive** grace period. I also hardened the extracted version so the grace counter resets after recovery; using total poll count would make a healthy long-running restore fail on a late transient response. The corresponding restore polling changes can be dropped after merge. The command test root passes (88/88, no leaks).
- #569 — preload registry variants. Centralizes model-spec parsing and preserves suffixes such as `owner/model:i8` and `owner/model:gguf:Q4_K_M`, recognizing a backend prefix only when it is explicit and supported. This can land independently of eager-preload orchestration, and the corresponding parsing changes can be dropped after merge. Main-runtime and standalone-runtime coverage passes.
- #570 — HTTPx oversized-body handling. Returns 413 for body-size overflow on both initial and pipelined reads, before invoking the application handler. I added real-socket regression tests for both paths. The corresponding HTTPx change can be dropped after merge. The full HTTPx suite passes (504/504).

The contributor's authorship is retained on the extracted original commits. The small follow-up commits that add/harden regression coverage are attributed separately. None of these PRs depends on the larger architectural portions of #561.

For the remaining work in #561, I still recommend splitting by independently reviewable subsystem/behavior rather than asking reviewers to validate the mega-PR as one unit. These five removals should make that next split materially smaller.

ajroetker:
[codex] Review findings for UX, correctness, efficiency, and performance

I reviewed the current PR head (`7f3ecdb4`) against current `main`. I think this still needs changes before merge. The focused restore/Raft/metadata/table-write suites pass, but the following concurrency, secure-cluster, and scale cases are not covered by those tests.

1. **[P1] Restore-job persistence uses a linearizable read as a write receipt**

   In `zig/pkg/antfly/src/metadata/server.zig` (`put`, `delete`, and `deleteMany`, around lines 737–773), the code proposes a transition, calls `ensureLinearizableRead()`, and then checks the projection. A ReadIndex is tied to the committed index when the read begins; it does not prove that the immediately preceding proposal was committed or applied. In a multi-voter cluster this can report `RestoreJobCommitNotApplied` even though the accepted proposal commits shortly afterward.

   Use `proposeWithReceipt()` and wait until the returned Raft index is applied, while checking leadership/term. Then optionally verify the projected value. A deterministic three-voter test should hold the proposal uncommitted while the read request is created.

2. **[P1] Secure follower forwarding for table create/drop omits internal authentication**

   `MetadataHttpService.metadataClientForNode()` in `zig/pkg/antfly/src/metadata/service.zig` (around lines 4776–4781) constructs a raw client without `withInternalServiceAuth()`. The new create/drop routes use this when the receiving metadata node is a follower. In an authenticated cluster, behavior therefore depends on whether the load balancer selects the leader or a follower.

   Centralize authenticated construction in a shared metadata `httpClient()` helper and add an auth-enforced multi-node test for follower forwarding.

3. **[P1] Distributed-transaction candidate fallback also bypasses its authenticated client helper**

   `beginGroupAtNode()` and `prepareGroupAtNode()` in `zig/pkg/antfly/src/api/distributed_txn.zig` (around lines 303–371) construct raw `ApiHttpClient`s. The type already has an authenticated `httpClient()` helper around line 210. Use that helper so leadership-churn recovery works in secure clusters, and test a forced first-candidate miss with internal auth enabled.

4. **[P1] Native restore staging blocks reads for the entire table**

   `reconcileReplicaRootTablesWithWriteCache()` in `zig/pkg/antfly/src/api/table_writes.zig` (starting around line 9169) begins a read-blocking group refresh activity before download, snapshot materialization, and runtime/index repair. Read admission treats any read-blocking group operation as table-wide blocking, so one slow shard restore can make the whole table unavailable even though the old published generation remains valid.

   Stage under the existing read-compatible group-operation mechanism and acquire the exclusive/read-draining fence only for final publication. Add a test that blocks staging and verifies reads continue against the old generation until atomic cutover.

5. **[P1] A transient checkpoint persistence failure can strand a running restore**

   `RestoreJobWork.run()` in `zig/pkg/antfly/src/api/http_server.zig` (around lines 13922–13955) calls `requeuePending()` after fenced/persistence errors, but `requeuePending()` in `zig/pkg/antfly/src/api/restore_jobs.zig` only inserts jobs whose phase is already `.queued`. Once `begin()` has changed the job to `.running`, this becomes a no-op: the worker exits, its scheduled marker disappears, and the supervisor has no pending job to rediscover until a leadership rebuild or restart.

   Distinguish ownership loss from a transient failure while still leader. Ownership loss can be left for the new leader; a current leader needs a durable `running -> queued` retry transition with backoff, or bounded retained retry work if persistence itself is unavailable. Add a no-leadership-change recovery test.

6. **[P2] Restore-progress change detection is quadratic on every store heartbeat**

   `localRestoreProgressNeedsScan()` in `zig/pkg/antfly/src/data/runtime.zig` (around lines 4616–4634) loops over local groups while repeatedly linearly searching ranges, progress entries, and group IDs. At 4,096 groups/ranges, a no-restore heartbeat can do roughly 16 million range comparisons.

   Build per-snapshot indexes once (local group set, range by group ID, progress by table/node/group) so the scan is linear, and skip it entirely when there are no active restore intents.

7. **[P2] Restore-progress publication performs serial per-shard HTTP and Raft work**

   `reportLocalRestoreProgress()` in `zig/pkg/antfly/src/data/runtime.zig` (around lines 14709–14755) sends each upsert/delete sequentially. Each record also fetches metadata `/head` before its mutation, and each mutation becomes its own metadata proposal. A restore with thousands of shards therefore causes approximately two serialized HTTP round trips and one Raft proposal per progress record.

   Add a bounded batch API/transition command—similar to the existing `reportStoreStatuses()` pattern—with one identity check per batch and a record/byte-size cap (for example, 128–256 records). Combine this with the indexed change detector above.

The first five are correctness/availability issues I would treat as merge blockers. The final two should be addressed before this becomes the production path for large provisioned tables.

ajroetker:
[codex] Existing abstractions to reuse and suggested PR split

Several parts of this PR can be simplified by using patterns that already exist in the API/runtime:

| Problem | Existing building block | Recommended use |
|---|---|---|
| Exact replicated write completion | `proposeWithReceipt()` in core/runtime/host Raft layers | Wait for the returned index to be applied; do not use a ReadIndex as the write receipt |
| Linearizable reads | `ensureLinearizableRead()` / linearizable head and snapshot operations | Keep these for read fencing only |
| Durable metadata mutation loops | Existing replace-table and cutover methods that propose, drive progress, and poll projection | Consolidate behind a generic receipt-based `propose + wait-for-applied-index` helper |
| Internal authentication | Existing `httpClient()` helpers and `withInternalServiceAuth()` | Eliminate raw clients in forwarding/fallback paths |
| Restore read availability | `tryBeginReadCompatibleGroupOperation()` and read-compatible transition activity | Use during download/staging/repair; fence reads only for publication |
| Bulk metadata updates | `reportStoreStatuses()`-style bounded reporting | Batch restore-progress upserts/deletes and validate identity once per batch |

In particular, we do already have “linear”/linearizable operations, but `ensureLinearizableRead()` is not sufficient for the restore-job write path. `proposeWithReceipt()` is the correct starting point because it binds the wait to the exact accepted proposal. I would elevate that into a shared production metadata helper returning or waiting on `{ term, index }`, with leadership/deadline checks, and have the domain methods verify their expected projection afterward.

This PR is still large enough that I recommend a dependency-ordered stack:

1. **Authenticated metadata and transaction forwarding**
   - Table create/drop follower forwarding.
   - Transaction candidate fallback.
   - Shared authenticated client constructors.

2. **Raft progress and exact-applied metadata mutation primitive**
   - Progress-only round plumbing.
   - `proposeWithReceipt()` plumbing and shared wait-for-applied-index helper.

3. **Schema-before-index-load correctness**
   - Schema propagation through managed reopen/load call sites.
   - This is independently understandable and testable. (`db.zig` itself is no longer changed in the current diff.)

4. **Durable data-Raft replay identity and checkpoint retry**
   - Replay identity coverage/pruning.
   - Apply checkpoint retry and state-machine retirement behavior.

5. **Replicated restore-job ownership and backoff**
   - Durable job store, leader rebuild, supervisor/scheduler.
   - Include the exact-write receipt and stuck-running-job fixes.

6. **Restore-progress protocol and batching**
   - Metadata representation/validation.
   - Bounded batch transport plus indexed change detection.

7. **Native generation restore handoff/convergence**
   - Download/staging, runtime reconciliation, and atomic publication.
   - Include read-compatible staging and the final short publication fence.

I would keep implementation and its tests together in each PR. Because the HTTP/server files overlap heavily, these should be stacked sequentially (or rebased as a patch stack) rather than opened as independent branches that will conflict. This structure leaves the final PR focused on the restore handoff itself instead of mixing routing, authentication, Raft mechanics, job scheduling, progress transport, replay identity, and storage publication.

ajroetker:
[codex] Extraction PRs are now open and validated:

- #576 — authenticated transaction participant leader rediscovery (main-based)
- #577 — progress-only Raft runtime rounds (main-based)
- #578 — exact metadata proposal receipts/application waits (stacked on #577)
- #579 — schema-before-managed-index load ordering (main-based)
- #580 — retry-safe durable data-Raft replay/checkpoints and replica retirement (stacked on #579)
- #581 — authenticated, at-most-once metadata table mutation forwarding (stacked on #578)
- #582 — exact restore-job worker ownership and durable bounded requeue/backoff (stacked on #578)

These include the review fixes rather than copying the unsafe forms from this PR: service auth is retained on fallback/forwarded requests, mutation ambiguity is not blindly replayed, metadata waits use the exact proposal receipt, the new regressions are wired into real focused targets, replay retries are term/index checked, and stale restore owners cannot requeue a newer attempt.

What intentionally remains in #561 is the coupled restore-progress protocol plus native-generation handoff/convergence. That is now the core feature rather than another “small extraction”: publishing it separately in its current form would preserve the quadratic progress scans, per-record HTTP/Raft serialization, and long read-blocking restore fence called out above. It should be revised around a bounded batch progress mutation/index and a read-compatible staging phase with a short exclusive publication fence before landing.

ajroetker:
[codex] Follow-up on metadata table mutation forwarding

The current extraction plan chose the wrong abstraction boundary. Table create/drop is a multi-step metadata workflow, so raw Raft follower proposal forwarding cannot directly provide the synchronous contract: a proposal forwarded by a follower has no local `{term,index}` acceptance receipt, and continuing the workflow from that follower would mix forwarded proposals with potentially stale local projections.

However, #581 overcorrects by adding a parallel application-level forwarding stack: custom metadata leader resolution, a new retry loop, dedicated create/drop forwarding endpoints, and separate protocol-marker/outcome handling. We already have the mature application-level Raft forwarding behavior in the data path: bounded deadlines and hops, campaign ownership, authenticated forwarding, leader rediscovery/fallback, and explicit not-proposed/unknown/committed outcomes. That implementation is currently coupled to data batch payloads, but its routing and outcome contract is the abstraction to generalize.

I am replacing that approach with a new focused PR based on the #577 stack:

1. Extract a group-agnostic Raft mutation forwarding context/policy while preserving the current data-batch wire contract.
2. Reuse its deadline, hop, authentication, leader fallback, and outcome classifications for metadata mutations.
3. Carry table create/drop in a metadata operation envelope with a stable operation/idempotency identity.
4. Admit a durable orchestration intent before performing the leader-owned multi-step workflow, so transport ambiguity converges by observing the operation instead of blindly replaying it.
5. Bind completion to exact proposal receipts/projection state rather than treating a linearizable read as a write receipt.

Raw Raft forwarding alone is insufficient for the present multi-step workflow, but bespoke metadata forwarding is unnecessary duplication. The replacement PR will supersede #581 rather than landing its current architecture.

ajroetker:
[codex] The replacement design is implemented in #584.

It generalizes the existing data-Raft forwarding budget/hop/campaign/outcome policy, uses one authenticated metadata mutation envelope with a stable operation identity, and makes table definition plus range publication/removal one atomic metadata transition. Completion is tied to the exact proposal receipt and verified projection. #581 has been closed as superseded.

Focused metadata service, metadata logic, data runtime, and real multi-node forwarding simulations are green (426 tests across the four targets).

PR #574, opened 2026-08-26
fix(pdf): harden native JBIG2 and JPEG 2000 image decoding

ajroetker:
## Summary
- add a bounded pure-Zig arithmetic JBIG2 decoder for the segment profile used by PDF image masks (symbol dictionaries, refinement/aggregation, text regions, and generic regions)
- decode JBIG2Decode streams and optional JBIG2Globals in the native PDF image path
- preserve mask failures instead of silently dropping pixel-defining transparency
- keep platform renderers unchanged

## Regression
The FRASER Treasury Bulletin page 25 now renders through the native Zig backend at 1196x1600 / 150 DPI with both dense numeric tables and rules intact:
https://fraser.stlouisfed.org/title/treasury-bulletin-407/february-1939-6505?page=25

The source stream exercises two arithmetic symbol dictionaries (including refinement aggregation), an immediate text region, and an immediate generic region. The decoded 2393x3201 mask contains 509,397 black pixels and the final native page render matches the correct reference composition.

## Verification
- zig build lib-pdf-test --summary all (223/223)
- zig build lib-pdf-safety-test --summary all (3/3)
- zig build lib-image-jpeg2000-test --summary all (247/247)
- direct native render of the reported Treasury page at 150 DPI


dovinmu:
```
 PR #574 was built at commit 3b17b6291ef29b054c421dceb1ad6a99cdebed4f and tested against a deterministic 1% OfficeQA Pro sample: 7 of 697 Treasury Bulletin PDFs, totaling 1,053 pages. All seven contain JBIG2 streams, with
 1,039 /JBIG2Decode tokens across the sample.

 Document ingestion completed successfully: all 1,053 pages were extracted, 2,560 chunks were indexed, and full-text/vector queries worked. However, OCR fallback attempted 148 page renders and failed on 68; 59 failures
 across six PDFs were InvalidJbig2SymbolCount from the new native JBIG2 decoder. This means #574 advances the previous failure from “JBIG2 unsupported” to “JBIG2 decode attempted but rejected,” but does not yet cover the
 OfficeQA scan profile.

 A strong reproducer is:

 - PDF: https://fraser.stlouisfed.org/files/docs/publications/tbulletin/1985_03_treasurybulletin.pdf
 - SHA-256: 6fcfabf9f21f9e8bc0565eed1f9430ecc18c543ee540e066ed24329e69e04b70
 - 238 pages extracted and 468 chunks produced
 - 36 OCR attempts; 21 failed with InvalidJbig2SymbolCount
 - Failing pages include 2, 3, 4, 5, 6, 8, 10, 14, and 16

 The error can originate from several symbol-dictionary consistency checks currently sharing the same error code, so the next step is to isolate one failing page, distinguish the exact failing check, and retain its JBIG2
 stream as a regression fixture.
```

PR #581, opened 2026-08-27
fix(metadata): forward table mutations securely

ajroetker:
Extracted from #561 and stacked on #578 (`codex/metadata-proposal-receipts`).

## What this changes

- routes public table create/drop requests received by a metadata follower to the observed metadata leader
- forwards canonical JSON bodies so valid table names are not constrained by URL path-segment syntax
- signs follower-to-leader requests with the existing internal service-auth mechanism
- preserves at-most-once semantics: only a protocol-marked, pre-admission `NotLeader` proof is replayed; transport and post-admission authority ambiguity surface as `MetadataMutationOutcomeUnknown`
- bounds table names and create bodies before expensive parsing or routing work
- keeps public HTTP errors actionable (`400`, `409`, `413`, `405`) without converting ambiguous outcomes into false success/failure

## Validation

- `zig build lib-metadata-service-test --summary all` (61 passed, including all forwarding, auth-header, ambiguity, validation, and leader-rediscovery regressions)
- `zig build api-http-runtime-test --summary all` (44 passed)

The PR deliberately uses the existing metadata HTTP client, authority classifier, and service-auth signer instead of adding another transport path.


ajroetker:
[codex] Superseded by #584.

#584 retains the secure follower-routing fixes from this PR but replaces the bespoke create/drop forwarding stack with the shared data/metadata Raft mutation policy, one operation envelope, and an atomic table-topology transition tied to an exact proposal receipt. This avoids running a multi-proposal workflow across follower projections and removes the parallel routing/outcome contract called out in the #561 follow-up.

PR #596, opened 2026-08-29
Fix/distributed restart restore

excenter:
## Why

Sequential member restarts followed by backup, drop, and restore exposed several
places where a replicated table did not converge even though the individual
components appeared healthy.

This PR primarily addresses that distributed recovery path. It also includes two
independent fixes discovered while building and operating the same PR environment.
Those are called out separately below.

## Primary thread: replicated restart and restore

- Own nested admin snapshot values before releasing temporary snapshots.
- Order metadata snapshots by their applied index so a metadata restart cannot
  leave data processes rejecting newer catalog state.
- Release read, write, and runtime-status storage owners when replicas retire.
- Keep published replica catalog changes after directory sync reports uncertain
  durability.
- Admit derived-index backlog on the leader before proposing replicated writes.
- Retry metadata requests during active table generation replacement.
- Keep linearizable read polling from advancing Raft time outside the managed
  progress driver.
- Derive fresh restore identities from the durable restore job.
- Keep restored placements unavailable until every voter reports complete,
  matching repair.
- Route table create, table drop, and restore progress through authenticated
  metadata-leader routes. Retry only responses proving the mutation was not
  admitted.
- Register focused Zig regression gates.

The new opt-in e2e scenario starts three metadata and three data processes with
replication factor three. It writes a fixed corpus, sequentially restarts the
metadata leader and every remaining member, checks voter agreement, durable node
and store identities, applied-index catch-up, and matching documents on every
replica, then backs up, drops, restores with fresh identities, repairs followers,
and verifies a post-restore replicated write.

## Separate thread: ARM64 release build scheduling

Production ReleaseFast measurements reached 8.96 GB for `api_kernel` and more
than 18 GiB for `distributed`, while their Linux reservations remained at 6 GiB
and 8 GiB. Raise those reservations to 10 GiB and 24 GiB so the build scheduler
does not admit combinations that cannot fit and turn completed work into an
out-of-memory retry.

This changes build scheduling only; it does not change the resulting binaries.

## Separate thread: Kubernetes structured Events

The operator's generated Role granted Event writes only through the legacy core
API group. Add matching `events.k8s.io` permissions, keep the controller marker
and generated Role aligned, and test the structured Event rule.

This is independent of replicated table recovery.

## Validation

- Dedicated ARM64 production build passed.
- Three-metadata, three-data restart and restore scenario passed.
- Zig formatting passed.
- Go formatting passed.
- Python formatting and lint checks passed, excluding existing upstream `B008`
  findings.
- The replicated e2e scenario collects successfully.

## Related upstream work

While this investigation was underway, upstream independently landed fixes in
several of the same restart and recovery paths:

- `6628c7a86` — own remote metadata snapshot data
- `5ffe4067a` — restore persisted Raft membership on restart
- `f5fc8280d` — add progress-only Raft runtime rounds
- `fa2631e02` and `ae4cc2322` — make data apply replay retry-safe and validate
  replay completion
- `b7a77c511` through `6c7ea6bcf` — recover and fence restore attempts

This work independently found bugs related to those commits while exercising the
complete three-metadata, three-data restart and restore path. During the rebase,
duplicate implementations were removed in favor of the upstream versions. This
PR retains the remaining non-overlapping fixes, integration wiring, and coverage
needed for the full sequence to converge.


excenter:
note AI was used to help me zig where I'd previously go'd, and this pr is far from perfect. if all that is taken is meaningful test cases to pre-validate the 3x3 runtime I'll be happy. thanks for your consideration.

ajroetker:
fwiw the raft semantics here are super tricky, I'm working on adding the necessary raft changes https://github.com/antflydb/antfly/pull/584

ajroetker:
thanks again for the work you've done on testing and implementation!!

ajroetker:
I've fixed the e2e suite to have a test for 3x3 scenarios specifically and did some work on raft in https://github.com/antflydb/antfly/pull/584 so this should all be good on main now!!

Issue #604, opened 2026-08-30
Track antfly-storage-kernel compiler VMA pressure and build-memory reductions


ajroetker:
## Summary

Tagged release builds of `antfly-storage-kernel` are failing inside LLVM with `Buffer allocation failed`. The immediate Linux failure is virtual-memory-area exhaustion in the official statically linked Zig host compiler, not exhaustion of the runner memory limit.

Main currently cross-compiles the Linux AMD64, Linux ARM64, and Darwin ARM64 release archives on Linux, so this host-compiler behavior can affect the whole release matrix.

## Evidence

- The 32 GiB publish pod peaked around 15.5 GB and recorded zero cgroup OOM or OOM-kill events.
- The host had the default `vm.max_map_count=65530`. GKE Autopilot exposes this as read-only to the runner pod, so CI cannot raise it with `sysctl`.
- The failing static compiler made 1,173,021 successful `mmap` calls and 1,162,280 successful `munmap` calls during the traced build.
- The terminal sequence failed tiny operations, including a 152 KiB `mmap`, with `ENOMEM`. Linux can also return `ENOMEM` from `munmap` when splitting a mapping would exceed the process map limit.
- Disabling `-ffunction-sections` and `-fdata-sections` changed peak compiler RSS by only about 23 MiB and failed the same way.
- Whole-kernel `ReleaseSmall` lowered sampled compiler RSS by about 0.9 GiB and completed LLVM, but it is not an ideal blanket change because it changes runtime optimization.

Diagnostic runs:

- Allocation trace and sectioning A/B: https://github.com/antflydb/antfly/actions/runs/33327060072
- Confirmation that the runner ceiling is 65,530 and cannot be changed in Autopilot: https://github.com/antflydb/antfly/actions/runs/33331940581
- Successful glibc-linked Zig control: https://github.com/antflydb/antfly/actions/runs/33332235847

## Successful control

Using the exact Zig 0.16.0 package from pinned Nixpkgs changes only the Linux host compiler from the official static musl executable to a dynamically linked glibc executable. The release target remains musl or Darwin and `ReleaseFast` remains unchanged.

The exact ARM64-musl `ReleaseFast` storage-kernel build succeeded with:

- status: 0
- peak compiler RSS: 7,998,896 KiB (7.63 GiB)
- peak pod usage: 14,104,858,624 bytes (13.14 GiB)
- peak compiler VMA count: 155
- unchanged host map ceiling: 65,530
[...]

PR #605, opened 2026-08-30
ci: use glibc-linked Zig for release builds

ajroetker:
## Summary

- install Zig 0.16.0 from a pinned Nixpkgs revision for tagged release archives
- apply the same host-compiler change to container runtime artifact builds
- verify both the Zig version and that the selected compiler is dynamically linked
- keep all release targets and `ReleaseFast` settings unchanged

## Why

The official Linux Zig tarball is statically linked and the `antfly-storage-kernel` build reaches Linux `vm.max_map_count=65530` before it reaches the 32 GiB runner memory limit. The terminal LLVM allocation failure was a 152 KiB `mmap`, while cgroup OOM counters remained zero.

The glibc-linked Zig control completed the exact ARM64-musl ReleaseFast build with a peak of 155 VMAs, 7.63 GiB compiler RSS, and 13.14 GiB total pod usage. Since main cross-compiles all tagged release targets on Linux, the host-compiler change covers the Linux and Darwin release archives without changing their target ABI.

Tracking issue: #604

## Validation

- `actionlint .github/workflows/antfly-release.yml .github/workflows/antfly-container.yml`
- `git diff --check`
- successful exact storage-kernel probe: https://github.com/antflydb/antfly/actions/runs/33332235847

The Nix installer action and Nixpkgs revision are SHA-pinned, and the workflow rejects a compiler whose version is not exactly 0.16.0 or whose executable is not dynamically linked.

### DIFF
--- a/.github/workflows/antfly-container.yml
+++ b/.github/workflows/antfly-container.yml
@@ -33,6 +33,9 @@ env:
   CLOUD_BUILD_WORKER_POOL: projects/antfly-image-artifacts/locations/us-central1/workerPools/antfly-container-builders
   ZIG_ARTIFACT_BUCKET: antfly-image-artifacts-zig-build-artifacts
   ZIG_VERSION: 0.16.0
+  # Keep the glibc-linked host compiler reproducible while continuing to emit
+  # static musl release artifacts for both architectures.
+  NIXPKGS_REV: c5c4a43b0e8056328ec4529f735cabdb8f1942bb
 
 jobs:
   build-artifact:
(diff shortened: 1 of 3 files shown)


ajroetker:
Darwin ARM64 release validation passed on this PR: https://github.com/antflydb/antfly/actions/runs/33335497623

- exact production archive command: `aarch64-macos`, `ReleaseFast`, Metal and system BLAS enabled
- build elapsed: 32m50.60s
- maximum process RSS reported by `/usr/bin/time`: 8,034,096 KiB (7.66 GiB)
- live glibc Zig sample: 121 VMAs; pod peak 16,365,895,680 bytes (15.24 GiB)
- executable and C API library both verified as Mach-O 64-bit ARM64
- `libantfly.dylib`: 18,497,632 bytes, below the 20 MiB retained-code budget
- archive and diagnostic artifact upload succeeded

PR #609, opened 2026-08-31
Build production CUDA inference image

bpopadiuk:
Codex (GPT-5): Adds the reproducible image prerequisite for the production scale-to-zero GPU inference service.

## Summary

- add a manual, `main`-only workflow that builds an amd64 glibc Antfly runtime with the CUDA `sm89` artifact
- package a digest-pinned CUDA 13.0 runtime image with ClipClap and GLiNER2 Q4_K models baked into `/models`
- publish only the full-commit-SHA GAR tag, resolve its immutable digest, sign that digest, and print the deployment input
- keep the normal multi-architecture Antfly runtime image and all production configuration unchanged

## Validation

- `actionlint .github/workflows/antfly-inference-gpu-container.yml`
- `bash -n scripts/packaging/build_zig_release_archive.sh`
- packaging flag help and invalid-value checks
- `docker buildx build --check --platform=linux/amd64 -f zig/Dockerfile.inference-gpu .`
- exact `x86_64-linux-gnu` ReleaseFast build and Linux CUDA artifact inspection completed during preparation (`sm89`, `cubin`, `sm_89`)

The image workflow itself repeats the CUDA artifact identity checks before publishing. It does not create or enable the production Cloud Run service.

### DIFF
--- a/.github/workflows/antfly-inference-gpu-container.yml
+++ b/.github/workflows/antfly-inference-gpu-container.yml
@@ -1,0 +1,143 @@
+name: CUDA inference container
+
+on:
+  workflow_dispatch:
+
+env:
+  FORCE_JAVASCRIPT_ACTIONS_TO_NODE24: true
+  GCP_REGION: us-central1
+  GCP_PROJECT: antfly-image-artifacts
+  GCP_REPOSITORY: containers
+  GAR_REGISTRY: us-central1-docker.pkg.dev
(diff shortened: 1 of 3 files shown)


bpopadiuk:
Codex (GPT-5): Closing this in favor of validating the normal Antfly container on a production L4 first. Linux release artifacts already compile the CUDA backend and dynamically load the Cloud Run-provided driver; model delivery will be handled as a separate concern. A specialized image should only be reconsidered if the normal-image GPU smoke test demonstrates a concrete gap.

PR #636, opened 2026-09-04
release: add trusted maintenance line policy

ajroetker:
## Summary

- add a controller-owned release-line map (`0.2 -> v0.2.x`, `0.3 -> main`) and derive source refs exclusively from canonical tags
- replace direct tag-triggered builds with an approved main-owned release-cut workflow that snapshots the selected branch tip, writes an immutable request, and creates the annotated tag last
- carry release line, source ref, and validated ref head through schema-v5 build/promotion provenance while retaining schema-v4 recovery compatibility
- codify main/v*.x branch protection and immutable v* tag rulesets, audit them at privileged release/GC entry points, and enable CI on maintenance PRs
- document the backport-only maintenance lifecycle once main advances to 0.3

## Verification

- `scripts/release/test.sh` (10 packaging tests, 111 release tests)
- `actionlint` on all changed workflows
- `scripts/release/validate_workflow_actions.py`
- `git diff --check`

## Rollout after merge

1. Synchronize this final pre-0.3 main state into `v0.2.x`, so the maintenance branch has the CI trigger/support contract.
2. Apply `scripts/release/github_environment.py` to create the protected `release-cut` environment.
3. Apply `scripts/release/github_rulesets.py` to protect `main`, `v*.x`, and `v*` tags.
4. Cut subsequent releases with `gh workflow run antfly-release-build.yml --ref main -f tag=v0.2.1-rc.N`.

The live ruleset audit currently reports both managed rulesets as missing; this is intentional until the introducing PR and final maintenance-branch synchronization are merged.

### DIFF
--- a/.github/workflows/antfly-artifact-build.yml
+++ b/.github/workflows/antfly-artifact-build.yml
@@ -86,12 +86,16 @@ jobs:
           [[ "$SOURCE_REF_HEAD" =~ ^[0-9a-f]{40}$ ]]
           if [ "$RELEASE_CHANNEL" = nightly ]; then
             resolved="$(python scripts/release/release_lines.py nightly)"
+            test "$RELEASE_LINE" = "$(jq -r .release_line <<<"$resolved")"
+            test "$SOURCE_REF" = "$(jq -r .source_ref <<<"$resolved")"
           else
-            resolved="$(python scripts/release/release_lines.py resolve --tag "$RELEASE_TAG")"
+            python scripts/release/release_lines.py verify-provenance \
+              --tag "$RELEASE_TAG" \
+              --release-line "$RELEASE_LINE" \
+              --source-ref "$SOURCE_REF" \
+              --allow-closed >/dev/null
             test "$SOURCE_COMMIT" = "$SOURCE_REF_HEAD"
           fi
-          test "$RELEASE_LINE" = "$(jq -r .release_line <<<"$resolved")"
-          test "$SOURCE_REF" = "$(jq -r .source_ref <<<"$resolved")"
           git fetch --no-tags origin "$SOURCE_REF"
           current_source_head="$(git rev-parse FETCH_HEAD)"
           git merge-base --is-ancestor "$SOURCE_REF_HEAD" "$current_source_head"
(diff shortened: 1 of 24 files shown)


ajroetker:
Post-merge rollout/status update

#636 merged as `a17c806e`, was reverted by #638 as `be5f5f4d`, and #639 is currently open as the revert-revert while we revisit the design.

No rollout steps from this PR were applied before the revert:

- `v0.2.x` remains at `bb8a284d` and does not contain #636.
- The `release-cut` environment does not exist.
- The repository currently has no rulesets.
- Neither `main` nor `v0.2.x` currently has branch protection.
- No release tags were changed as part of this rollout.

For the current phase, `main` still owns the 0.2 line. We do not need a `0.2 -> v0.2.x` mapping yet; `0.2 -> main` preserves current behavior. The handoff only becomes necessary when `main` moves to 0.3.

Smaller proposed design

1. Keep the existing manually pushed `v*` tag request workflow. Do not add a release-cut environment or a workflow that creates tags.
2. Change only the default-branch release controller so it derives the trusted source branch from the tag release line.
3. Configure `0.2 -> main` now. At the 0.3 version bump, change that to `0.2 -> v0.2.x` and add `0.3 -> main`.
4. Require the tag commit to equal the selected branch tip and carry that selected ref/head through provenance.
5. Keep privileged build/promotion policy loaded from `main`; do not let a maintenance-branch workflow self-authorize a release.

This cannot be implemented only in `v0.2.x` while retaining the current trust boundary: the `workflow_run` controller and privileged policy are loaded from the default branch. The narrow fix belongs in the main-owned controller, while the release source may come from `v0.2.x` after handoff.

Green CI enforcement is separate from release-line ownership. We can protect against branch deletion and force-pushes, and optionally require PRs/reviews, without adding `required_status_checks`. GitHub settings can be changed directly and do not intrinsically require a PR; a small separate PR is useful only if we want the desired ruleset recorded and auditable in the repository.

Recommendation: do not merge #639 as-is. Replace it with the smaller controller-only release-line change, and handle branch/tag protection policy separately after agreeing on enforcement.

PR #689, opened 2026-09-09
fix(inference): align operator CLI and runtime model configuration

bpopadiuk:
## Summary

Fix standalone inference configuration so the operator can recover v0.2.1 deployments independently of a runtime release and use the same generated command with the next main-based runtime.

Previously, model pulls wrote to `/models`, but `antfly inference run --config` in v0.2.1 ignored the configured model directory and preload list. Successful downloads could therefore leave inference permanently unready.

## Operator compatibility

- Resolve nested-over-flat overrides, defaults, model kinds, and artifact references before generating JSON and CLI arguments. Explicit empty preload lists and zero/unbounded model limits are preserved.
- Pass v0.2.1-supported model flags, including explicitly configured `--ml-dir`. Pullers, the inference server, and shared model mounts use the same resolved model directory. Argument changes participate in the existing rollout hash.
- Emit legacy-compatible JSON: move nested model fields to the flat standalone spelling, remove an emptied `inference` object, and retain other nested settings with the legacy schema's required client URL field. Keep admission settings and per-model policies.
- Treat empty/null optional backend, format, and quantization fields as unspecified in both outputs.
- Require recognized task hints for generated eager preloads, including the default loading strategy and per-model eager overrides. Missing hints no longer silently classify embedding models as generators. Users can instead supply an explicit `spec.config.preload` kind or use lazy/bounded discovery.
- Report invalid runtime configuration through `ConfigurationValid=False` and the degraded pool phase before creating/updating workloads. Revalidate unchanged generations on operator upgrades; correcting the task hint restores reconciliation.

## Runtime configuration

- Read `models_dir`, `ml_dir`, `max_loaded_models`, and `preload` from flat or nested `--config`, with nested fields winning individually and no remote `api_url` required.
- CLI flags override config regardless of argument order. CLI preloads select the list; exact kind/reference/backend matches retain config-only residency and memory policies. Ambiguous duplicate matches are rejected.
- Normalize format/quantization preferences into the model reference used for resolution, preserving artifact tags and rejecting conflicting selections. Reuse the raw preload parser and fix cleanup of partially parsed entries.

## Validation

- Full operator tests pass with Kubernetes envtest, along with `go build ./...`, `make lint` (zero issues), workflow `actionlint`, and `git diff --check`.
- Twelve identical reconciler-generated CPU contract scenarios pass against both v0.2.1 and the freshly built PR runtime (18 seconds and 21 seconds locally). They cover eager/lazy loading, missing task hints with an explicit eager kind, empty/null optional fields, nested overrides, tagged GGUF models, omitted config kinds, and a missing-directory negative control. Tests require warming/readiness and real finite, nonzero embeddings.
- The same Go contract is wired against the freshly built candidate in inference E2E CI. Operator-only changes trigger candidate E2E. A separate Python suite covers config-only startup and CLI precedence.
- Focused runtime/config tests cover model identity and policy matching, precedence, tagged references, malformed settings, and allocation cleanup.

## Release scope

The operator can ship separately for v0.2.1; config-only per-model policies require the new runtime. Automatic eager-kind detection from downloaded metadata remains a future runtime capability. Existing eager pools without recognized task hints need `tasks` or an explicit preload kind.

This does not add standalone config support for other settings such as `keep_alive_ms`. CPU process tests do not replace container, live Kubernetes, or GPU/A4B validation. No live deployment is included.

### DIFF
--- a/.github/workflows/antfly-operator-go.yml
+++ b/.github/workflows/antfly-operator-go.yml
@@ -97,7 +97,7 @@ jobs:
           tar -xzf "${runtime_dir}/${archive}" -C "$runtime_dir"
           ANTFLY_RUNTIME_BIN="${runtime_dir}/antfly" GOWORK=off go test \
             -tags runtimeintegration ./controllers/inference \
-            -run '^TestInferenceRuntimeContract$' -count=1 -v -timeout=15m
+            -run '^TestInferenceRuntimeContract$' -count=1 -v -timeout=20m
 
       - name: Run lint
         run: make lint
--- a/.github/workflows/antfly-operator-go.yml
+++ b/.github/workflows/antfly-operator-go.yml
@@ -41,6 +41,7 @@ env:
 jobs:
   test:
     runs-on: arc-antfly-standard
+    timeout-minutes: 30
     defaults:
       run:
         working-directory: go/pkg/operator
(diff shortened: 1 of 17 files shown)


.github/workflows/antfly-operator-go.yml
```diff
@@ -77,6 +78,27 @@ jobs:
       - name: Run tests
         run: make test
 
+      # This is a required step, not an env-gated unit test: execute the
+      # reconciler's generated command against the supported release on CPU.
+      # Update the version AND checksum when advancing the runtime baseline.
+      - name: Test generated inference command against released runtime
```
ajroetker, line 84: hmm maybe this should be in the zig e2e tests? so main also works

dovinmu (approved):

PR #778, opened 2026-09-17
graph agent queries

huckleberryseed:
This adds a new graph query type `graph_agent`. This type of query allows an agent to walk a path along the graph, beginning at some arbitrary starting node. We feed the data from that node as instructions to our agent along with our query. At each step the agent chooses the next best node to path to, accumulating additional instructions and data from the next node in its path.

Queries are structured like this:
```
{
  "full_text_search": {
    "query": "<natural-language starting search>"
  },
  "graph_queries": {
    "<query-name>": {
      "index": "<graph-index-name>",
      "graph_agent": {
        "start": {
          "result_ref": "$query_results",
          "limit": 1
        },
        "query": "<user request>",
        "instruction_field": "<document field containing node instructions>",
        "max_steps": 4,
        "neighbor_limit": 8,
        "instruction": "<general rules for how the agent should reason and respond>",
        "generator": {
          "provider": "<provider>",
          "model": "<provider-model-id>",
          "api_key": "${secret:<secret-name>}",
          "temperature": 0,
          "max_tokens": 160
        },
        "include_trace": true
      }
    }
  },
  "limit": 10
}
```

We perform any existing sort of antfly supported query to find a starting node. Once we determine that node, we begin executing down a path of connected nodes accumulating results and further instructions as we go.  

Responses look like this:
```
{
  "kind": "agent",
  "final_key": "<final-node-key>",
  "answer": "<final answer>",
  "steps": [
    {
      "from_key": "<starting-node>",
      "selected_key": "<second-node>",
      "model_output": "{\"done\":false,\"next_key\":\"<second-node>\",\"answer\":null}"
    },
    {
      "from_key": "<second-node>",
      "selected_key": "<third-node>",
      "model_output": "{\"done\":false,\"next_key\":\"<third-node>\",\"answer\":null}"
    },
    {
      "from_key": "<third-node>",
      "selected_key": null,
      "model_output": "{\"done\":true,\"next_key\":null,\"answer\":\"<final answer>\"}"
    }
  ]
}
```

We get a full execution chain of the path taken and intermediate results for each turn.


ajroetker:
[codex]

I recommend handling model-directed graph navigation in the agents API as a retrieval strategy, and addressing the issues below before merging.

Reviewed commit `13f8068d66967438cd88e335a72a18daca19e259`. This is a static review; tests were not run.

The overlap with retrieval's tree search is substantial, but the policies differ: existing tree retrieval translates navigation into canonical graph traversals, ranks branches, and expands promising branches while accumulating evidence. This PR calls a model at each node to select one neighbor and optionally interpret node fields as instructions. That is a useful new policy, but it should share the existing agent orchestration rather than introduce a separate agent lifecycle inside the query HTTP handler.

Findings:

- **P1: Row-level authorization is bypassed.** `executePublicGraphAgentQuery` explicitly discards `row_filter_json`; node lookup and neighbor queries go directly through `source`. A user with table-read permission but restricted rows could have excluded documents sent to the model. The normal query handler injects those filters and attaches graph authorization. All start-node resolution, node reads, and neighbor reads need the same authorization context. [HTTP implementation](https://github.com/antflydb/antfly/blob/13f8068d66967438cd88e335a72a18daca19e259/zig/pkg/antfly/src/api/http_server.zig#L12045)
- **P1: The response does not match the query contract.** The handler returns the bare `{kind:"agent", ...}` object instead of placing it under the named `graph_results` entry in a query response. It collects `agent_name` but never uses it, and the PR does not add an agent variant to `GraphResult`. Typed clients and consumers expecting the normal query response cannot consume this as advertised. [Response construction](https://github.com/antflydb/antfly/blob/13f8068d66967438cd88e335a72a18daca19e259/zig/pkg/antfly/src/api/http_server.zig#L12150)
- **P2: Instructions and evidence do not accumulate.** Each generation receives only the original request, current document, and current neighbors. Previous node instructions, documents, and model outputs are not supplied to later decisions, contrary to the behavior described in the PR. A later node cannot reliably answer using information encountered earlier in the path. [Agent loop](https://github.com/antflydb/antfly/blob/13f8068d66967438cd88e335a72a18daca19e259/zig/pkg/antfly/src/api/graph_agent.zig#L210)
- **P2: `direction` and `edge_types` are accepted but ignored.** Neither is carried into the runner or its neighbor traversal, so requests can walk the wrong relationships. Forward these constraints into every neighbor query. [Neighbor construction](https://github.com/antflydb/antfly/blob/13f8068d66967438cd88e335a72a18daca19e259/zig/pkg/antfly/src/api/http_server.zig#L12081)

Recommended API and implementation direction:

1. Keep graph reads in the query API and put model-directed navigation in retrieval orchestration. Add an explicit graph-navigation configuration alongside `tree_search`, retaining its single-path policy and optional node instructions.
2. Execute node and neighbor reads through the existing authorized query path, preserving row filters and the request's security context.
3. Reuse the agent's iteration/context budgets, accumulated evidence, steps, usage reporting, and answer generation.
4. Share traversal helpers between tree search and graph navigation while keeping their selection policies distinct. Existing heuristic branch expansion and model-selected single-path navigation need not become the same algorithm.
5. Keep node instructions explicit opt-in behavior: following a stored workflow and retrieving document evidence have different semantics.

This preserves the useful new navigation capability while avoiding a second public agent lifecycle inside `graph_queries`.

ajroetker:
Thanks for the contributions! These have been great, I think this is a great feature and overlaps heavily with the /agents api that already exists and the retrieval agent endpoint specifically so I had codex go ahead and review the PR with some of that extra context. Let me know what you think! I can push up follow-up for these or PR to this branch or let you handle them :) if you haven't started on them tomorrow I can pick them up!

huckleberryseed:
I'm fine with it living under the `/agents` api. The main thing I'm wanting is the ability to treat data I'm getting at each node in the search as further instructions for the agent. Currently I can only accumulate additional information to answer the original query.

ajroetker:
[codex]

Your clarification—that data encountered at each node should become further instructions for the agent, rather than only additional evidence for the original question—is preserved in the implementation merged in #779.

The capability now lives at `/db/v1/agents/retrieval`, with navigation policy under `steps.retrieval.navigation`. For your single-path workflow, the configuration is:

```json
{
  "query_index": 0,
  "strategy": "graph",
  "selection": "agentic",
  "index": "workflow",
  "instruction": "Follow the workflow, applying instructions from each visited node.",
  "instruction_field": "instructions",
  "max_steps": 8
}
```

The ordinary query in `queries[0]` finds the starting node; an explicit `start_key` can select it directly. Each visited node's configured field is presented as an opted-in workflow instruction, so it can guide subsequent navigation and the answer. Earlier instructions and evidence remain in the same conversation through the final answer. The model chooses among offered connected neighbors, and every read retains the caller's filters and authorization. The enclosing retrieval agent supplies the generator, iteration/context budgets, and optional generation step.

A few boundaries to make explicit:

- `instruction_field` names a top-level string field; include it if the query uses a `fields` projection. Other document fields and unvisited neighbor documents remain evidence, not instructions.
- Node instructions can guide navigation and answer generation within the caller's permissions and budgets; they do not enable arbitrary external actions.
- Results use the retrieval agent's `hits`, navigation `steps`, and `generation`, rather than reproducing the original raw `model_output`/`final_key` response shape.
- Regression tests verify instruction delivery and retention with a mocked generator. A real-model workflow evaluation demonstrating behavior changes from later-node instructions remains a validation gap.

The related tree exploration capability uses `strategy: "tree"`; your workflow use case remains the distinct `graph` + `agentic` path.

See the [design document](https://github.com/antflydb/antfly/blob/main/docs/design/retrieval-navigation.md) and [request examples](https://github.com/antflydb/antfly/blob/main/docs/guides/graph-navigation.mdx).

Closing this PR as superseded by the now-merged #779. Thank you for contributing the original implementation and clarifying the instruction-driven workflow use case.

PR #780, opened 2026-09-17
docs: fix orphaned restart and readiness notes in quickstart

shindakun:
The model-install step told readers to restart the process without naming it. It now says restart `antfly standalone`, matching the same advice later on the page.

The index readiness note sat two sections before any index existed, and named a state that does not exist. It moves to the `index wait` step, and `queryable` becomes `queryable_partial`.


ajroetker:
Thanks @shindakun <3

Issue #181, opened 2026-06-06
v0.2.0-rc.2: embeddings backfill with missing model — ModelNotFound classified as non-retryable, no recovery on restart, but any new write heals it; POST /indexes over existing docs returns 500


markhayden:
Found during v0.2.0-rc.2 smoke testing. Verified against the official
  `antfly_0.2.0-rc.2_Darwin_arm64.tar.gz` release artifact (`antfly 0.2.0-rc.2 (zig runtime)`),
  single-node `swarm` mode, darwin-arm64, auth off, clean data dir, **empty models dir**.
  Part of a series: #TBD (create-time schema), #TBD (FTS `_all`), #TBD (CLIP pull), #TBD (dimension docs).

  Setup: managed dense index
  `{"name":"emb","type":"embeddings","field":"title","dimension":384,"embedder":{"provider":"antfly","model":"BAAI/bge-small-en-v1.5"}}`
  with the model not on disk.

  ## Observed behavior

  1. **Index created together with the table, then insert a doc** →
     `backfill_state: "failed"`, `error_count: 1`, `retryable_error_count: 0`,
     `fatal_error_count: 1`. Log: `enrichment worker failed: ModelNotFound`.
     No attempt to lazy-download from HuggingFace. (Lazy *load from disk* works fine — the first
     use after a pull logs `selected backend metal for <models-dir>/BAAI/bge-small-en-v1.5`.)

  2. **After `antfly inference pull BAAI/bge-small-en-v1.5`:**
     - passive wait → stays `failed`
     - **full server restart → stays `failed`** (failed backfill is not retried on open)
     - **one new batch write → `failed` → `ready`**, and the *previously failed* documents are
       embedded too: `total_indexed` covers all docs, and `semantic_search` returns the doc that
       was written while the model was missing.

  3. **Different creation path, harder failure:** adding the same index to a table that already
     has documents (`POST /db/v1/tables/{t}/indexes/emb`) with the model missing →
     `500 index create failed`, and the index is **not created at all** (synchronous backfill
     failure aborts index creation rather than leaving a failed-but-existing index).

  ## Why this seems wrong

  The write-path heal proves `ModelNotFound` is retryable — the failed work is fully recoverable
  in-process once the model exists — yet it's counted in `fatal_error_count` with
  `retryable_error_count: 0`, and nothing retries it without user traffic. Operators who pull the
  missing model and then check index status (or restart the node) see a permanently-failed index
[...]

PR #785, opened 2026-09-17
fix(antfarm): expire the connections cache so pulled models appear

timkaye11:
## Problem

Antfarm's connections cache stored no timestamp, and the mount effect returned
early whenever an entry existed — so the expanded model inventory was fetched
once per page load and never again. A model installed with `antfly inference
pull` stayed invisible in the dashboard until a full browser reload.

The server was never at fault. Model listing (`server.zig:18605`) and
resolution (`server.zig:9374`) each build a fresh `ModelRegistry` and walk the
models directory per request, and `ModelRegistry` holds no cache — a pulled
model is servable immediately. That is why restarting `antfly standalone`
appeared to help: the page reload that came with it cleared the browser-side
cache.

Affected surfaces: every model dropdown and index form — `ModelsPage`,
`ConnectionsPage`, `IndexForm`, `GeneratorSelector`, `use-inference-registry`,
and the ten playground pages.

## Fix

A 30s cache window with background revalidation while a model view is open:

- Background refreshes never set the loading or error state, so the page no
  longer drops to a skeleton, and a failed revalidation keeps the last good
  inventory instead of showing a full-page error.
- State seeds from a cached entry regardless of age (stale-while-revalidate),
  so a remount renders the previous list instead of blanking.
- Cache writes broadcast to every consumer of the key, and a tick whose entry
  is still fresh does no work: N mounted widgets cost one request per window
  rather than N, cannot disagree about which models exist, and all recover
  together from a transient failure.
- The poll does not force a refetch. Forcing bypasses the server-side cache
  that keeps dashboards off remote provider listing APIs; only an explicit
  Retry does that.
- A forced refresh is authoritative and is never overwritten by a concurrent
  cache-backed read, which would otherwise make Retry visibly do nothing.

## Docs

The quickstart, inference, and multimodal guides blamed the running process and
told readers to restart `antfly standalone`. That advice has no mechanism
behind it, and in the one real failure in this area (#181) a full server
restart explicitly left the failure in place while a single write healed it.
Replaced with what actually happens.

Note this rewords the Callouts added in #780.

## Testing

- 12 new tests, 19 total in the hook suite.
- Every guard is mutation-verified: reverting any one of them individually
  makes tests fail.
- The bug reproduces on `origin/main`: 6 of the new tests fail against its
  version of the hook.
- biome clean, `tsc --noEmit` 0 errors, quickstart doc validator and release
  scope pass.
- The embedded dashboard bundle is regenerated, and `diff -qr
  apps/antfarm/dist ../zig/pkg/antfly/antfarm` (the SDK CI check) passes from a
  clean rebuild of the committed tree.
- Playwright is unaffected: its two active specs cover `/` and the Tables page,
  neither of which mounts a connections consumer.

No runtime changes — Zig source is untouched; the only files under `zig/` are
the regenerated dashboard assets.


ts/apps/antfarm/src/hooks/use-connections.ts
```diff
@@ -114,20 +190,47 @@ function useConnectionsInternal(includeModels: boolean): ConnectionsState {
...
+    if (!includeModels) return;
+
+    const revalidate = () => {
+      if (document.visibilityState === "hidden") return;
+      const entry = connectionsCache.get(cacheKey);
+      // Another consumer of this key already refreshed inside the window.
+      if (entry && isFresh(entry)) return;
+      // A forced refresh is already fetching authoritative data for this key.
+      if (connectionsInFlight.has(`${cacheKey}|refresh=true`)) return;
+      void fetchConnections(undefined, { background: true });
```
ajroetker: [codex] **[P2] Invalidate background polls when the API endpoint changes**

If endpoint A's background poll is pending when the user switches to endpoint B, A's response can arrive after B loads and replace B's inventory. This call passes no abort signal, and the mount effect resets `isMountedRef.current` to `true` for B, so the old response passes the guard in `fetchConnections` and updates the current consumer with A's models. The view then shows models from the wrong endpoint until a subsequent refresh.

Give this polling effect an abort controller, pass its signal to `fetchConnections`, and abort it during cleanup, or validate the request's endpoint before applying its result. A regression test that starts A's poll, switches to B, resolves B, then resolves A reproduces the issue; it passes with the cleanup fix.

ajroetker (commented):
[codex] Found one P2 issue: an in-flight background poll can overwrite the model inventory after switching API endpoints.

Validation: all 19 existing hook tests pass. An additional endpoint-switch regression test fails on this commit and passes when the polling effect uses an abort signal that is cancelled during cleanup. Validation ran in an isolated temporary directory.

ajroetker (commented):
[codex] Fresh review of e632e4a: the TypeScript endpoint-switch fix is correct, but the shipped dashboard bundle still contains the previous implementation. One P2 finding below.

Validation: all 20 existing hook tests pass, plus an additional polling lifecycle test covering deduplication, hidden tabs, visibility recovery, and cleanup. No additional source-code findings. Tests ran in an isolated temporary directory.

zig/pkg/antfly/antfarm/index.html
```diff
@@ -11,7 +11,7 @@
             href="https://fonts.googleapis.com/css2?family=Inter:wght@400;500;600;700&family=Roboto+Mono:wght@400;500;700&family=Silkscreen:wght@400;700&display=swap"
             rel="stylesheet"
         />
-      <script type="module" crossorigin src="/assets/index-Bsx3V2hN.js"></script>
+      <script type="module" crossorigin src="/assets/index-BogcqQkn.js"></script>
```
ajroetker: [codex] **[P2] Regenerate the shipped dashboard bundle after the endpoint-switch fix**

The bundle referenced here (`assets/index-BogcqQkn.js`, line 102) is byte-for-byte identical to the previously reviewed revision (916d2fe) and still invokes background polling as `h(void 0,{background:!0})`, with no abort signal or abort during effect cleanup. The TypeScript fix therefore does not reach users of the packaged dashboard: `zig/build.zig` installs `pkg/antfly/antfarm` into `share/antfly/antfarm`, and standalone serves those assets. Switching API endpoints while a poll is pending can still replace the new endpoint's model inventory with the old endpoint's response. Rebuild and commit the dashboard assets, including the updated script reference in `index.html`, so the shipped code includes the fix.

